Issue #2, opened 2019-02-19
lexer: long strings and long comments


hishamhm:
Currently, the lexer does not support long strings (`[[hello]]`, `[===[hello]===]`) and long comments (`--[[hello]]`, `--[===[hello]===]`).

Since the implementation of those would be rather similar, I listed both here. The lexer needs to remember the number of `=` seen once it enters a long string or comment and only close it when it sees the correct number.

hishamhm:
Fixed by #5!

Issue #1, opened 2019-02-19
lexer: single quoted strings


hishamhm:
Currently the lexer does not support `'single quoted strings'`, only `"double quoted strings"`

hishamhm:
Fixed by #5!

PR #5, opened 2019-02-22
#1 and #2

kipras:


hishamhm:
Cool! Nice that the new code does not add type errors to the count in `tl tl.tl` too :)

Since this new code is not exercised by `tl tl.tl`, would you mind adding some minor test for long comments and single quotes under `spec/` just so that when we run `make` it will catch any regression?

kipras:
Yes, i'm gonna start adding test coverage maybe not just for this but for other parts of lexer parser as well (if you don't mind), that's in my todo list.

I think a simple unobtrusive test (where you don't dive deep down in resulting lexer/parser data structures (so there shouldn't be any risk of breaking tests with actually working code (which is perhaps the only reason why you _might_ want to avoid writing tests for alpha stage code, i.e. the fact that the structure may change substantially)) is simply parsing a tl program and then exporting the resulting Lua (and maybe also running it) and checking if it matches what you expect.

But for things like long comment matching - to fully test this it requires that comments be preserved in exported Lua code, which is what i was going to do next. After that i'll add tests for this.

hishamhm:
Preserving comments tends to complicate the lexer/parser pair, because a
comment can appear essentially anywhere (and one would need to correctly
track its position to put it above or beside a statement, inside an
expression or argument list, etc etc). I would avoid this for now. As for
the unobtrusive test, I agree: simply testing that the presence of a valid
long comment produces no errors and that an invalid comment produces an
error, something like that.

hishamhm:
@kipras when commenting here, I totally missed out the fact that I had _not_ pushed the `spec/` folder with my initial minimal tests! (It must have sounded weird suggesting to add a test when there's no testsuite around)

hishamhm:
@kipras anyway, this is merged! I also made a lot of progress in my long flight from POA to SFO — we're down to 2 errors now! The first one sounds like it should be fixable (for the second one the code is doing something really dynamic with Lua tables)

kipras:
Great. I'll add the tests for this pr when i can, been a bit swamped..

PR #9, opened 2019-03-06
Tests for parsing strings, long strings and long comments

kipras:
Finally had some time to add tests for strings and comments


tl.tl
```diff
@@ -259,6 +259,11 @@ function tl.lex(input: string): {Token}
          end
       end
    end
+
+   if #tokens > 0 and tokens[#tokens].tk == nil then
+      drop_token()
+   end
```
kipras: This here is so that we would not add an invalid (unfinished) last token, in case of a lexer error. Because then the parser would crash (because of missing `.tk` field).

Although it would be better to raise a lexer error here instead (as there shouldn't be unfinished tokens).

tl.tl
```diff
@@ -1275,6 +1280,7 @@ parse_statements = function(tokens: {Token}, i: number, errs: {ParseError}): num
 end
 
 function tl.parse_program(tokens: {Token}, errs: {ParseError}): number, Node
+   errs = errs or {}
```
kipras: This here is so tl would not crash if there are parse errors and you did not supply an `errs` parameter (tests don't do that). From what i understand the `errs` parameter is supposed to be optional?

hishamhm: Yeah, since this is the entry point of the parser it makes sense to keep this optional. :+1:

tl.tl
hishamhm: > Although it would be better to raise a lexer error here instead (as there shouldn't be unfinished tokens). Because now a program with some unfinished token (like an unclosed long string/comment) would be treated by tl as a valid tl program (and tl would only process the part of the program up until that unfinished token).

Agreed, it would be better to make it return nil+error and handle it in the proper places.

hishamhm:
Merged! :)

PR #17, opened 2019-11-03
A couple more tests (to check that filenames for errors are returned correctly from required files)

kipras:


spec/typecheck/typecheck_error_spec.lua
```diff
@@ -0,0 +1,73 @@
...
+   it("unknowns include filename", function ()
+      local tokens = tl.lex("local x: string = b")
+      local _, ast = tl.parse_program(tokens, {})
+      local errors, unknowns = tl.type_check(ast, true, "foo.tl")
+      assert.same("foo.tl", unknowns[1].filename, "unknowns should contain .filename property")
+   end)
+
+   it("unknowns in a required package include filename of required file", function ()
+      local io_open = io.open
+      io.open = function (filename, mode)
```
hishamhm: when you monkey-patch a function like this in a test, I think it's necessary to switch it back once the test is done, otherwise if a new test is added later to the file that assumes the function is the standard one, it will fail.

You can use a `finally(function() io.open = io_open end)` block to ensure the function is reverted to the original on both test success and failure

kipras: Good point, thanks, i'll fix this.

There's additional test sandboxing support in busted, called "insulation" [1], but that wouldn't help either in this case looks like it - it only sandboxes loaded packages `package.loaded` and globals, and in this case it is _modifying_ a package.

[1] https://olivinelabs.com/busted/#insulate-expose-blocks

kipras:
Fixed

hishamhm:
@kipras merged! :tada:

PR #18, opened 2019-11-07
top-level global declarations

hishamhm:
For declaring globals, should I be doing contortions to parse ambiguous syntax like

```
x: T
```

at the toplevel and detect it as a global declaration or should I just add a new `global` keyword for global declarations? This PR does the ambiguous-syntax handling option for toplevel globals, but it produces some weird error messages in case of misuse.

I think I'll drop this and go with the `global` keyword, but this is the code of the other version.

### DIFF
--- a/spec/declaration/global_spec.lua
+++ b/spec/declaration/global_spec.lua
@@ -1,0 +1,83 @@
+local tl = require("tl")
+
+describe("global", function()
+   describe("undeclared", function()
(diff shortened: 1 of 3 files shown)


hishamhm:
Tweeting about it: https://twitter.com/hisham_hm/status/1192234943658811393

hishamhm:
Retracting this in favor of #19

PR #22, opened 2019-11-21
Enums

hishamhm:
Many Lua APIs use magic strings where in C you'd use enums. Only a limited set of strings are valid, and they're usually used as literals passed as arguments to functions.

This PR adds support for `enum` type declarations, to let you check those literals for typos.

* A value of an `enum` type is a string that can only be assigned to a literal from a specific set.
* An `enum` can be assigned to a variable of type `string`.
* A `string` can only be assigned to a variable of an enum type if it is a literal declared in that an enum.
* As usual, if you force your way with a cast, all bets are off.

Enums with the same contents but different names are considered to be different (i.e. enums are checked against each other nominally, not structurally).

```lua
local Direction = enum
   "north"
   "south"
   "east"
   "west"
end

local function go(d: Direction)
   print("I am going " .. d .. "!") -- d works as a string!
end

go("west") -- life is peaceful there!

local d: Direction = "west"
go(d) -- in the open air!

go("rest") -- ERROR, not a Direction! caught your typo! ^_^

local d = "west" -- infers as string
go(d) -- ERROR, not smart enough to do flow-analysis!
go(d as Direction) -- compiler believes what you say!

go("we" .. "st") -- ERROR, not smart enough to do constant subexpression elimination!
```


kikito:
The "use integers for enums" pattern is also prevalent in Lua. For example, the `log_level` in [ngx.log](https://github.com/openresty/lua-nginx-module#ngxlog) (`ngx.LOG`, `ngx.WARN`, etc). I would venture they are more prevalent than string-based enums, given that Lua often has to deal with C-like enums.

I suggest taking [a page from typescript's book](https://www.typescriptlang.org/docs/handbook/enums.html#numeric-enums) and supporting both.

```
local DirectionInt = enum
  north = 1
  south = 2
  east = 3
  west = 4
end
local DirectionString = enum
  north = "north"
  south = "south"
  east = "east"
  west = "west"
end
```

In addition to Enums, Typescript has two more tools which could be used here: [String Literal Types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#string-literal-types) and [Union Types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#union-types). I think this is too much: it has three ways to very similar goals†

Which brings me to the second question: is some form of Union Type somewhere in the roadmap?
In rust, [there is no difference between enums and union types](https://doc.rust-lang.org/book/ch06-01-defining-an-enum.html)‡. I think that is a good design decision. I understand that this Pull is about Enums and not Union Types, but I suggest keeping them in mind. For examplethis rust enum encodes several ways of interacting with a program in a single type:

```
enum Message {
    Quit,
    Move { x: i32, y: i32 },
    Write(String),
    ChangeColor(i32, i32, i32),
}
```
I think this kind of thing would be difficult to replicate without some sort of explicit separator (the comas, in rust) besides whitespace. Also, this can be further complicated if we introduce [the G word](https://doc.rust-lang.org/book/ch10-00-generics.html) :)

---
 
† Typescript's String Literal types would not satisfy `Enums with the same contents but different names are considered to be different`, they are compared structurally, and can be used "anonymously", right there in a function call definition.

‡ rust's enums variants are, internally, integers. They can be [explicitly defined](https://doc.rust-lang.org/rust-by-example/custom_types/enum/c_like.html) but they cannot be strings. Makes sense since rust doesn't have interned strings.

hugomg:
Will the type system try to recognize a series of if-elseif statements over an enum variable? For example, to warn if you forget to handle one of the cases?

Full-blown unions would certainly be a nice-to-have, but the tricky bit is figuring out how to deal with the elimination forms (if-elseif, pattern matching, etc). In Typed Lua the type system needs to be flow-aware to be able to understand when it can narrow the type of a variable because of an if statement.

Another minor question is what should the compiler do if it encounters a typo like `if direction == `"wst"`? Should it be a compile-time error or should it silently perform a string comparison?

hishamhm:
> The "use integers for enums" pattern is also prevalent in Lua. For example, the log_level in ngx.log (ngx.LOG, ngx.WARN, etc). I would venture they are more prevalent than string-based enums, given that Lua often has to deal with C-like enums.

@kikito That is already covered by records: `ngx.WARN` is just a field of type `number` in the `ngx` record. If you define the record `ngx`, then a mistake and type `ngx.WARNING`, tl will already warn you.

>  north = "north"
>  south = "south"
>  east = "east"
>  west = "west"

No way in hell I'm going to add a new "pattern" based on repetitive typing :) I want to _get rid_ of all those stupid `local print = print` lines with an optimizer pass eventually :)

> Which brings me to the second question: is some form of Union Type somewhere in the roadmap?

That's a great question. In fact @hugomg asked me the same thing over email this week. :) I'm trying to see how far I can go without it, because the more recursive types you add, the more complicated things get really fast. For now it's easy to avoid more complex structures since I'm mostly typechecking things I wrote myself so I got more careful as soon as tl started to typecheck itself (it's amazing how fast you end up with constructs of complex types when you're happily coding away at a dynamic language, as I said in my FOSDEM talk), but typing other real-world Lua code might force my hand into adding more complex types. (You'll be happy to know that one of the first third-party libraries that I tried typechecking was your `inspect` — and yeah, it does some hard-to-type stuff, I had to put a bunch of casts in particular places!)

The one external library I'm _having_ to type with what I have right now is the Lua standard library itself. For now, I'm getting by with a restricted form of intersection types for functions only, which I don't even have syntax for declaring yet (so for now it's a privileged type for the stdlib only). It might allow me to get away without union types for longer. The thing about union types is that once you have them, you also need something in the language to be able to resolve them at compile time (like `match` in Rust or `case` in Haskell). And I know people are going to look at me weird when I add a new control structure to the language (and no, I don't plan to try to magically detect `if type(var) == "string"` like Typed Lua tried to do). I don't want to talk vaporware but I do have some obvious simplifications of union types in mind such as nullability checks.

So tl;dr: yes, I think about union types a lot and I know how Rust tries to keep everything nice and tidy by being very orthogonal, but for now I think special cases — i.e. a "larger" language with a few constrained features instead of a few very-general very-powerful mechanisms that interact with each other in explosive ways — will actually keep the language and its implementation simpler.

hishamhm:
> Will the type system try to recognize a series of if-elseif statements over an enum variable? For example, to warn if you forget to handle one of the cases?

@hugomg you're describing a switch-case, not if-elseif :)

> In Typed Lua the type system needs to be flow-aware to be able to understand when it can narrow the type of a variable because of an if statement.

I know, I sent enough PRs to Typed Lua fixing corner cases of its flow analysis to become sufficiently wary of going down that rabbit hole before attacking more important things! :smile: 

> Another minor question is what should the compiler do if it encounters a typo like if direction == "wst"`? Should it be a compile-time error or should it silently perform a string comparison?

Take off your type-theorist hat and put back your programmer hat: what would _you_ like the behavior to be? :wink: I know which one _I'd_ like better!

kikito:
> You'll be happy to know that one of the first third-party libraries that I tried typechecking was your inspect — and yeah, it does some hard-to-type stuff, I had to put a bunch of casts in particular places!

I am honored and humbled 🙂

> a "larger" language with a few constrained features instead of a few very-general very-powerful mechanisms that interact with each other in explosive ways

This was surprising for me to read at first, but upon reflection, I think it might be more Lua-ish than my initial proposal, yes.

Maybe the enums can "cooperate" with some "other thing" in order to "form" of union types, similarly to how in Lua multiline strings "cooperate" with comment prefixes to "form" multiline comments.

hishamhm:
> This was surprising for me to read at first, but upon reflection, I think it might be more Lua-ish than my initial proposal, yes.

It was surprising for me to write it too, because it sounds like it goes against the whole "mechanisms not policies" mantra from Lua. And yes, strictly speaking it is. But there is an inherent contradiction between generality and simplicity when it comes to type systems. My take here is that remaining a small language that "fits in your head" and being driven by solving practical problems are two more important design directives of Lua than mechanisms-not-policies. Lua does cut corners in conceptual purity when it comes to remaining practical and small (the `#` operator is probably the most infamous example: its semantics sound terrible when explained, but work well in practice and allow for a small and efficient implementation).

It may be possible to come up with a very small and very general type system for Lua that ticks all boxes of safety, size and practicality, but I'm not smart enough to come up with that so I'm coming up with what I can. :) Plus, even Lua itself wasn't as polished as it is in its early days and had its own design become more orthogonal over time. But it has to start somewhere!

hishamhm:
Well, since reactions to this were generally positive, I'll going ahead and merge this! Thank you @kikito and @hugomg for the feedback, keep it coming!

PR #41, opened 2020-02-01
Allow preloading modules

pdesaulniers:
Initilal work on https://github.com/hishamhm/tl/issues/35.

We can determine which modules to pre-load by setting the `preload_modules` key inside `tlconfig.lua`:
```lua
return {
	-- Execute the equivalent of `require('modulename')` before executing the script.
	preload_modules = {
		"definitions.love"
	}
}
```

I have not implemented the `-lmodule` command-line option yet.

### DIFF
--- a/spec/preload/preload_spec.lua
+++ b/spec/preload/preload_spec.lua
@@ -1,0 +1,85 @@
+local tl = require("tl")
+local util = require("spec.util")
+
+describe("preload", function()
+   it("exports global types", function ()
+      -- ok
+      util.mock_io(finally, {
+         ["love.d.tl"] = [[
+            global love_graphics = record
+               print: function(text: string)
+            end
+
+            global love = record
+               graphics: love_graphics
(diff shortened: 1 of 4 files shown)


hishamhm:
@pdesaulniers Nice! I merged this as-is as is definitely an incremental improvement. Noted about the `FIXME`, yeah, it would be cool to typecheck `tlconfig` using tl itself. But this PR is a good start!

About the `-lmodule` option, it can be added in a different PR — I've been trying to keep tl self-contained with no dependencies (I'm on the fence about keeping the `compat53` stuff, there are pros and cons), but I suppose at one point we'll need an actual argument parser! (One option would be to keep `tl.tl` dependency-free to make `tl.lua` easy to embed in Lua projects, but use dependencies in the `tl` CLI tool for convenience).

pdesaulniers:
OK, I'll look into integrating `argparse`.

PR #42, opened 2020-02-02
Integrate the 'argparse' library; add -l command-line option

pdesaulniers:
Implements the rest of https://github.com/hishamhm/tl/issues/35.

Note that running `tl script.tl` will now result in an error.

~~Also, it would be nice if we had tests for the CLI tool :)~~

### DIFF
--- a/tl
+++ b/tl
@@ -13,6 +13,10 @@ local function starts_with(str, start)
    return str:sub(1, #start) == start
 end
 
+local function trim(str)
+   return str:gsub("^%s*(.-)%s*$", "%1")
+end
+
 local function die(msg)
    printerr(msg)
    os.exit(1)
@@ -113,8 +117,19 @@ local tlconfig = get_config()
 
 local cmd = args["command"] or "check"
 
-if args["preload"] then
-   tlconfig["preload_modules"] = args["preload"]
+for _, preload_module_cli in ipairs(args["preload"]) do
+   local found_in_config = false
+
+   for _, preload_module_config in ipairs(tlconfig.preload_modules) do
+      if trim(preload_module_cli) == trim(preload_module_config) then
+         found_in_config = true
+         break;
+      end
+   end
+
+   if not found_in_config then
+      table.insert(tlconfig.preload_modules, preload_module_cli)
+   end
 end
 
 local function report_errors(category, errors)
@@ -143,7 +158,7 @@ local function report_type_errors(result)
 end
 
 for _, filename in ipairs(args["script"]) do
-   local result, err = tl.process(filename, nil, nil, nil, tlconfig["preload_modules"])
+   local result, err = tl.process(filename, nil, nil, nil, tlconfig.preload_modules)
    if err then
       die(err)
    end
--- a/tl
+++ b/tl
@@ -108,7 +108,9 @@ if not success then
       parser = get_parser(false) -- Reset the parser
       args = parser:parse()
    else
-      print(err)
+      print(parser:get_usage())
+      print()
+      print(string.format("Error: %s", err))
       os.exit(1)
    end
 end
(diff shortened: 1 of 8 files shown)


tl
```diff
@@ -50,36 +59,33 @@ end
...
+run_command:argument("script", "The tl script."):args(1)
+
+local args = parser:parse()
+
+local tlconfig = get_config()
+
+local cmd = args["command"]
+
+if args["preload"] then
+   tlconfig["preload_modules"] = args["preload"]
```
hishamhm: This is _replacing_ the list in tlconfig, right? I think it should probably add to it instead.

tl
```diff
@@ -107,70 +113,55 @@ local function report_type_errors(result)
...
-   print("       to run " .. filename .. " as a program")
-   print()
-   print("   tl gen " .. filename)
-   print()
-   print("       to generate " .. lua_name)
-   os.exit(0)
+      print("========================================")
+      print("Type checked " .. filename)
+      print("0 errors detected")
+      print()
```
hishamhm: I know these instructions you removed are sort of ugly, but I think at this early stage it's good to intstruct newcomers to ease the learning curve.

tl
```diff
@@ -50,36 +59,33 @@ end
...
-   run = true,
-   check = true,
-   gen = true,
-}
-
-if is_cmd[arg[1]] then
-   cmd = arg[1]
-   filename = arg[2]
-else
-   cmd = "check"
```
hishamhm: Is there a way in argparse to keep supporting `tl file.tl` without an explicit command?

To be honest, I don't know if it's just my habit of typing `tl tl.tl` speaking or if it's really something worth keeping because it's probably the first thing a user would try...

tl
pdesaulniers: This behavior is based on `tsc` https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#details

> Compiler options specified on the command line override those specified in the tsconfig.json file.

I don't mind changing it, though.

tl
pdesaulniers: Mmmh... It's kinda hacky, but we could call
```lua
local args, err = parser:pparse()
```
...and check if `err` starts with `unknown command`. If it does, we could assume that the 'unknown command' is actually a file.

Maybe there's a better way of doing this...

pdesaulniers: See https://github.com/hishamhm/tl/pull/42/commits/32b777e32ddde7af051d434cd5f861caaf33c42d.

tl
pdesaulniers: OK, I brought them back.

tl
pdesaulniers: Should be fixed in https://github.com/hishamhm/tl/pull/42/commits/ffff7ef8c42ae42c84df009baf741e34c314572d.

tl
hishamhm: Thanks!

tl
hishamhm: Aw, yeah, that _is_ pretty hacky. And mixing argument 1 with commands and filenames is conceptually ugly, the more I think about it.

hishamhm: I think we could drop that and require `tl check file.tl` indeed.

hishamhm:
> Also, it would be nice if we had tests for the CLI tool :)

@pdesaulniers  Added them! :grin:

pdesaulniers:
Awesome! I added some tests for the `-l` argument too.

We'll need some tests for `tlconfig.lua` as well. To do this, I think we'll need to add a `-p <path to directory with tlconfig.lua>` argument to the CLI. I'll do this in a later PR.

hishamhm:
I think this is good to go — could you rebase this PR and fix the conflicts so it can be merged?

pdesaulniers:
Like this? :)

tl
hishamhm: > This behavior is based on tsc https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#details

Oh, I wasn't aware. Well, let's see how people react and if there's confusion, we can change it later if we find it necessary.

PR #50, opened 2020-02-07
Strip excess newlines

ashn-github:
Running tl on arbitrary lua files produced a large number of excess
newlines.
This commit runs the generated lua string though a regex substitution
that will replace occurrences of two or more newlines with two newlines.
The output is a bit denser and looks closer to human-written code.

### DIFF
--- a/tl.lua
+++ b/tl.lua
@@ -3,12 +3,6 @@ local assert = require('compat53.module').assert or assert; local io = require('
    ["type_check"] = nil,
 }
 
-
-
-
-
-
-
 local inspect = function(x)
    return tostring(x)
 end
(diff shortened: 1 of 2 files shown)


hishamhm:
@ashn-github the new lines are kept on purpose (in fact, the pretty-printer goes through some pains to preserve them). The goal is to keep the line numbers of the generated .lua file to match your input .tl file. This is done so that runtime errors produce backtraces with the correct line numbers.

hishamhm:
Since I expect the programmer writing a `.tl` to maintain that one as their point of reference (and not the generated `.lua` file, which is more of a build artifact), I think it's more important to keep the backtrace truthful to the input than to compress the output `.lua` file (there are projects out there to minify Lua code if one doesn't care about preserving backtraces).

So, I'm closing this PR for now since we won't be merging this, but thank you for the contribution! Do keep providing feedback on tl, it's much appreciated!

PR #63, opened 2020-03-01
union types, take 1

hishamhm:
Will write docs asap, please look at the tests for now!

TODO:

- [x] union type syntax ( `string | number`)
- [x] `is` operation for runtime matching (`x is number`)
- [x] fix codegen for non-primitive types
- [x] type narrowing in statements (`if  x is number then ... else ... end`)
- [x] type narrowing within expressions (`s = s is string and s:lower()`)

**Edit:** I think the implementation for this first approach towards union types is pretty much ok to merge, but I'd like to call out one important limitation (the text below needs to go into the documentation too):

Since code generation for the `is` operator used for discrimination of
union types translates into a runtime `type()` check, we can only
discriminates across primitive types and at most one table type.

This means that these are valid:

```
local a: string | number | MyRecord
local b: {boolean} | MyEnum
local c: number | {string:number}
```

but these are not valid:

```
local invalid1: MyRecord | MyOtherRecord
local invalid2: {string} | {number}
local invalid3: {string} | {string:string}
local invalid4: {string} | MyRecord
```

Also, since `is` checks for enums currently also translate into
`type()` checks, this means they are indistinguishable from strings
at runtime. So, for now this is also not accepted:

```
local invalid5: string | MyEnum
```

This restriction between strings and enums may be removed in the future.
The restriction on records may also be lifted in the future if we start
registering metatables for them (it may be that a future release may
start supporting unions of records _with_ metatables, if we add a form
of record constructors with metatables).

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -462,6 +462,83 @@ declarations and multiple returns:
 f: function(function():(number, number), number)
 ```
 
+## Union types
+
+The language supports a basic form of union types. You can register a type
+that is a logical "or" of multiple types: it will accept values from multiple
+types, and you can discriminate them at runtime.
+
+You can declare union types like this:
+
+```
+local a: string | number | MyRecord
+local b: {boolean} | MyEnum
+local c: number | {string:number}
+```
+
+To use a value of this type, you need to discriminate the variable, using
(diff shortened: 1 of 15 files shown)


Zash:
Looks nice!

PR #79, opened 2020-04-03
Add command-line flag to disable compat53 insertions

### LINKED ISSUE #78: Add a flag to disable compat53 code insertions
As a feature request can we get a way to disable the automatic compat53 shimming and assume 5.3 or the most recent support version? The added code like:
```lua
local assert = require('compat53.module').assert or assert; local pairs = require('compat53.module').pairs or pairs;
```
is breaking my code in fengari and I'd not like to add compat53 if the only supported version is 5.3 anyway.

25A0:
This PR adds a command-line flag to skip compat53 insertions.

The flag `skip_compat53`, that is already used in `TypeCheckOptions`, is now exposed as a command-line option. The flag can be set with `--skip-compat53` from the command line, and is then passed to `tl.process` in `env`.

Example:

```shell
% # This file uses table.unpack, which would usually trigger the compat53 inserts.
% cat test.tl
local t = {1, 2, 3, 4}
print(table.unpack(t))

% # When passing --skip-compat53, compat53 inserts are not added
% tl --skip-compat53 gen test.tl
Wrote: test.lua

% cat test.lua
local t = { [1] = 1, [2] = 2, [3] = 3, [4] = 4, }
print(table.unpack(t))

% # Without the flag, compat53 inserts are added
% tl gen test.tl
Wrote: test.lua

% cat test.lua
local table = require('compat53.module').table or table; local _tl_table_unpack = unpack or table.unpack; local t = { [1] = 1, [2] = 2, [3] = 3, [4] = 4, }
print(_tl_table_unpack(t))

```

I'm open to any form of feedback or suggestions :)

So far there's no option to set this flag from the tl loader, but as far as I can tell the compat53 insertions are never added with the loader.

Resolves #78

### DIFF
--- a/spec/cli/gen_spec.lua
+++ b/spec/cli/gen_spec.lua
@@ -77,8 +77,7 @@ describe("tl gen", function()
       end)
    end)
 
-   describe("with/without --skip-compat53", function()
-
+   describe("with --skip-compat53", function()
       it("does not add compat53 insertions", function()
          local name = util.write_tmp_file(finally, "test.tl", [[
             local t = {1, 2, 3, 4}
@@ -94,8 +93,9 @@ describe("tl gen", function()
             print(table.unpack(t))
          ]], util.read_file(lua_name))
       end)
+   end)
 
-
+   describe("without --skip-compat53", function()
       it("adds compat53 insertions by default", function()
          local name = util.write_tmp_file(finally, "test.tl", [[
             local t = {1, 2, 3, 4}
@@ -111,6 +111,6 @@ describe("tl gen", function()
             print(_tl_table_unpack(t))
          ]], util.read_file(lua_name))
       end)
-
    end)
+
 end)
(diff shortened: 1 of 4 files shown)


hishamhm:
@25A0 Looks great! Merged! :)

We might change the CLI a bit in the future to allow for different code generation modes (e.g. generate different code when targeting PUC Lua or LuaJIT variants), but for now this flag is definitely sufficient.

hishamhm:
> So far there's no option to set this flag from the tl loader, but as far as I can tell the compat53 insertions are never added with the loader.

ohh, good catch. The loader currently doesn't type check, it just strips the types (for speed), and the compat53 insertions are being done as an AST transformation at type check stage...

I'm thinking that instead of detecting at type check which compat-needing functions are being used, we could, when using compat53, just prepend a fixed "prelude" which loads it and just `local`-izes all compat53 variables. This way this could be done in the loader without type checking.

PR #81, opened 2020-04-04
Rename some instances of "tl" to "Teal" in tutorial

pdesaulniers:
The tutorial was still referring to Teal as "tl". This PR tries to fix this.

In some places, I tried to disambiguate between Teal (the language) and tl (the compiler). Here are some examples:

> Unlike in Lua, global variables in **Teal** need to be declared, because **tl** needs to know its type. It also allows **tl** to catch typos in variable names [...]

>  If you make assignments of conflicting types, **tl** will tell you in its error message [...]

> Or we can load the **tl** package loader, which will allow require to load .tl files directly

Maybe it looks a bit confusing...

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -1,9 +1,9 @@
-# Programming With Types in tl
+# Programming With Types in Teal
 
-## Welcome to tl!
+## Welcome to Teal!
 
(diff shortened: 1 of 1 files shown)


hishamhm:
@pdesaulniers ooh, good catch! Looks _much_ better now!

Issue #92, opened 2020-04-25
Allow setting the working directory on the command-line?


pdesaulniers:
Related: https://github.com/teal-language/vscode-teal/issues/9

I think it would be useful if `tl` allowed setting the working directory of the type-checker by passing a command-line argument.

Let's say we have a project with the following structure:
```
src/
  - main.tl
  - point.tl

tlconfig.lua
```

And `main.tl` looks like this:
```lua
local point = require("point")
```

Then, we could type-check `main.tl` by running the following command at the root of the project:
```
tl check --cwd=src/ src/main.tl
```
(Technically, we could even remove the `src/` prefix in `src/main.tl`, but that would mess up shell completion)

hishamhm:
This effectively means adding the src/ directory to the require search path (which is doable today with environment variables TL_PATH or LUA_PATH).

Instead of just a rootdir flag, it would be useful to have a more general argument that adds (prepends?) dirs to the module search path, adjusting the package.path and package.cpath accordingly (we could guess the cpath library extension by looking at existing entries). Ideally this flag could be set multiple times and added to tlconfig.lua.

PR #94, opened 2020-04-25
Add option for prepending directories to the module search path

pdesaulniers:
Initial work on https://github.com/teal-language/tl/issues/92.

This PR adds the `-I --include` option for prepending directories to `package.path`. It also adds an `include` key in `tlconfig.lua`.

### DIFF
--- a/spec/cli/include_spec.lua
+++ b/spec/cli/include_spec.lua
@@ -1,0 +1,17 @@
+local util = require("spec.util")
+
+describe("-I --include argument", function()
+   it("adds a directory to package.path", function()
+      local name = util.write_tmp_file(finally, "foo.tl", [[
+         require("add")
+         local x: number = add(1, 2)
+
+         assert(x == 3)
+      ]])
+
+      local pd = io.popen("./tl -I spec check " .. name, "r")
+      local output = pd:read("*a")
+      util.assert_popen_close(true, "exit", 0, pd:close())
+      assert.match("0 errors detected", output, 1, true)
+   end)
+end)
--- a/spec/cli/include_spec.lua
+++ b/spec/cli/include_spec.lua
@@ -16,16 +16,13 @@ describe("-I --include argument", function()
    end)
    it("adds a directory to package.cpath", function()
       local name = util.write_tmp_file(finally, "foo.lua", [[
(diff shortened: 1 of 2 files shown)


tl
```diff
@@ -18,10 +18,21 @@ local function die(msg)
    os.exit(1)
 end
 
+local function find_in_sequence(seq, value)
```
hishamhm, line 21: The PR looks great but the indentation is a bit off? (4 spaces in a few places)

hishamhm:
I was thinking we could update `package.cpath` as well with `-I` (it could be useful for `tl run`) but that would be an "extra", I think functionality-wise this PR is good to go as is already!

tl
pdesaulniers: My bad! Should be fixed, I think.

pdesaulniers:
> I was thinking we could update package.cpath as well with -I

OK, I don't mind adding this right now. I went with `package.path` first since it's more useful for my use-case.

pdesaulniers:
It should now set `package.cpath` too.

spec/cli/include_spec.lua
```diff
@@ -0,0 +1,28 @@
...
+      local pd = io.popen("./tl -I spec check " .. name, "r")
+      local output = pd:read("*a")
+      util.assert_popen_close(true, "exit", 0, pd:close())
+      assert.match("0 errors detected", output, 1, true)
+   end)
+   it("adds a directory to package.cpath", function()
+      local name = util.write_tmp_file(finally, "foo.lua", [[
+         print(package.cpath)
+      ]])
+      local pd = io.popen("LUA_CPATH=\"/usr/lib/lua/?.so\" ./tl run -I spec/cli/ " .. name, "r")
```
hishamhm, line 21: I think this test will fail if people have things like LUA_CPATH_5_3 set in their environments (I do :grin: ). I believe it's okay to just do `print(package.cpath:match("spec/cli/") ~= nil)` above and check for `true` below (should also keep it crossplatform enough!)

pdesaulniers: Good call! Done.

hishamhm:
@pdesaulniers Merged! :grinning:

PR #95, opened 2020-04-28
tl run: support CLI arguments passed to the script

hishamhm:
`tl run hello.tl foo bar` did not pass `foo` and `bar` as arguments to `hello.tl` — this PR intends to fix that. :)

If my memory serves me right, Lua and LuaJIT have somewhat different behavior in their handling of `arg`. The invariants I kept here are:
* `arg[0]` is always the script name
* positive `arg` entries are the script arguments
* anything else (Lua interpreter, `tl` arguments) is in negative `arg` entries.

Our behavior is a little different from PUC Lua though, in that once `lua` sees a non-flag, it stops argument parsing immediately. argparse does not do that, it needs a `--` to stop interpreting arguments. I do some juggling of the arguments to preserve the invariants above.

### DIFF
--- a/spec/cli/run_spec.lua
+++ b/spec/cli/run_spec.lua
@@ -138,4 +138,143 @@ describe("tl run", function()
          assert.same("", output)
       end)
    end)
+
+   describe("with arguments", function()
+      it("passes arguments as arg", function()
+         local name = util.write_tmp_file(finally, "hello.tl", [[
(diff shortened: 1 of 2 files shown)


hishamhm:
@pdesaulniers I'm already using `-I .` in tests, the "harmless flag" came in handy for testing ;)

Issue #103, opened 2020-05-11
Project file for compilation options


iljio:
Hello,

I noticed that `tl` currently doesn't have a lot of options, like choosing an output directory or anything like that. Continuing that thought, I felt like the existence of further options might warrant a project file altogether, similar to `tsconfig.json` for TypeScript.

If I were to expect how such a project file would look, I would expect something along the lines of:
```
-- tlconfig.lua
return {
  directories = {
    input:    {"src/**/*.tl"},
    include:  {"lib/**/.*.lua"},
    output:   "dist"
  },
  flags = {
    transpiler = {
      save_sourcemaps = true,
      keep_comments   = true,
      check_requires  = true
    },
    warnings = {
      unused_variables  = true,
      unknown_variables = false
    },
    errors = {
      const_reassignment = true
    },
    compatibility = {
      version = "5.1"
    }
  },
  plugins = {
    "prettifier-5.1",
    "switch",
    "lambda"
  }
}
```

Of course, I'm just describing what it might look like, not implying that tl has to implement anything like plugins. (Although, the idea of a tl plugin sounds kind of neat...)

pdesaulniers:
Actually, `tl` already has support for a `tlconfig.lua` file, although it's not advertised anywhere :)

For the time being, these two keys are supported:

```lua
return {
	include = { -- add these directories to LUA_PATH and LUA_CPATH; matches the -I command line option
		"folder1/",
                "folder2/"
	},
        preload_modules = { -- execute the equivalent of require('modulename') before executing the tl script(s); matches the -l command line option
                "my.module",
                "my.other.module"
        }
}
```
I agree that `tlconfig.lua` (and the command line tool) should support something like the `input` and `output` keys in your example. I've opened some issues about this:

Input: https://github.com/teal-language/tl/issues/31
Output: https://github.com/teal-language/tl/issues/47

If ever you want to take a stab at implementing these features, you might want to look at this PR which adds support for the `include` key: https://github.com/teal-language/tl/pull/94

hishamhm:
> Actually, tl already has support for a tlconfig.lua file, although it's not advertised anywhere :)

Time for a `docs/tlconfig.md`? :grin:

pdesaulniers:
Instead, I think I would add `docs/compiler_options.md`, which would describe both `tlconfig.lua` and compiler switches.

hishamhm:
Since the project file already exists, I think we can close this one!

PR #104, opened 2020-05-12
Create compiler_options.md

pdesaulniers:
As mentioned in https://github.com/teal-language/tl/issues/103.

### DIFF
--- a/docs/compiler_options.md
+++ b/docs/compiler_options.md
@@ -25,3 +25,4 @@ return {
 | --- | --- | --- | --- |
 | `-l --preload` | `preload_modules` | `{string}` | Execute the equivalent of `require('modulename')` before executing the tl script(s). |
 | `-I --include` | `include` | `{string}` | Prepend this directory to the module search path.
+| `--skip-compat53` | | | Skip compat53 insertions.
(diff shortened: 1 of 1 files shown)


docs/compiler_options.md
```diff
@@ -0,0 +1,27 @@
...
+    }
+}
+```
+
+## List of compiler options
+
+| Command line option | Config key | Type | Description |
+| --- | --- | --- | --- |
+| `-l --preload` | `preload_modules` | `{string}` | Execute the equivalent of `require('modulename')` before executing the tl script(s). |
+| `-I --include` | `include` | `{string}` | Prepend this directory to the module search path.
```
hishamhm, line 27: Should this table include `--skip-compat53` as well? :thinking: 

(One thing we did at work was to generate the CLI reference doc out of the output of running `--help` on the CLI tool, just so we didn't miss any flags... for now we can maintain this file by hand, but eventually that's an idea worth exploring :) )

pdesaulniers: > One thing we did at work was to generate the CLI reference doc out of the output of running --help on the CLI tool, just so we didn't miss any flags

Yes, we could probably figure out some way to build this table programmatically.

Anyway, I've added the flag to the table.

PR #124, opened 2020-06-01
Tagged records

hishamhm:
Played a bit with this idea this weekend. 

The one big thing missing in this PR is actually checking that table literals conform to the tagged record type. That is, the example below works, but as you can see all literals at the bottom are using casts:

```lua
local WidgetType = enum
   "button"
   "field"
end

local Widget = record
   tag widget_type: string

   x: number
   y: number
end

local Button = record is Widget with widget_type = "button"
   text: string
end

local Field = record is Widget with widget_type = "field"
   width: number
   label: string
   input: string
end

local function draw(x: number, y: number, w: number, t: string)
end

local function show_widget(w: Widget)
   if w is Button then
      draw(w.x, w.y, 10, w.text)
   elseif w is Field then
      draw(w.x, w.y, w.width, w.label)
      draw(w.x, w.y, w.width, w.input)
   else
      draw(w.x, w.y, 10)
   end
end

local widgets: {Widget} = {
   { widget_type = "field", width = 20, label = "Name:", input = "" } as Widget,
   { widget_type = "field", width = 3, label = "Age:", input = "" } as Widget,
   { widget_type = "button", text = "OK" } as Widget,
   { widget_type = "button", text = "Cancel" } as Widget,
}

```

### DIFF
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -101,9 +101,15 @@ precedence, see below.
 
 *  typeargs ::= ‘<’ Name {‘,’ Name } ‘>’
 
-*  newtype ::= ‘record’ [typeargs] [‘{’ type ‘}’] {Name ‘=’ newtype} {Name ‘:’ type} ‘end’ |
+*  newtype ::= ‘record’ [typeargs] [‘is’ Name ‘with’ Name ‘=’ tagvalue]
+*                 [‘{’ type ‘}’]
+*                 {Name ‘=’ newtype}
+*                 { [‘tag’] Name ‘:’ type}
+*              ‘end’ |
 *      ‘enum’ {LiteralString} ‘end’ |
 *      ‘functiontype’ functiontype
+
+*  tagvalue ::= ‘false’ | ‘true’ | Numeral | LiteralString
 
 *  functiontype ::= [typeargs] ‘(’ partypelist ‘)’ [retlist]
 
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -400,6 +400,40 @@ local http = record
 end
 ```
 
+### Tagged records
+
+Teal supports a very simple form of subtyping for records: tagged records.
+You can specify a field of type number, string or boolean to act as a _tag_
+for a parent record type, and then define children record types that extend
+the parent record discriminating them based on the tag. This is how it
+works:
+
+```
+local Widget = record
+   tag widget_type: string
+
+   x: number
+   y: number
+end
+
+local Button = record is Widget with widget_type = "button"
+   text: string
+end
+
+local Field = record is Widget with widget_type = "field"
+   width: number
+   input: string
+end
+
+local function show_widget(w: Widget)
+   if w is Button then
+      draw_button(w.x, w.y, w.text)
+   elseif w is Field then
+      draw_field(w.x, w.y, w.width, w.input)
+   end
+end
+```
+
 ## Generics
 
 Teal supports a simple form of generics that is useful enough for dealing
--- a/spec/assignment/to_nominal_record_field_spec.lua
+++ b/spec/assignment/to_nominal_record_field_spec.lua
@@ -1,4 +1,5 @@
 local tl = require("tl")
+local util = require("spec.util")
 
 describe("assignment to nominal record field", function()
    it("passes", function()
@@ -32,4 +33,42 @@ describe("assignment to nominal record field", function()
       local errors = tl.type_check(ast)
       assert.match("in assignment: got number, expected Node", errors[1].msg, 1, true)
    end)
+end)
+
+describe("tagged records", function()
+   it("inherit fields from parent", util.check [[
+      local Node = record
+         tag kind: string
+         x: number
+         y: number
+      end
+
+      local EnumNode = record is Node with kind = "enum"
+         enumset: {string}
+      end
+
+      local e: EnumNode = {}
+      e.x = 12
+      e.y = 13
+      e.enumset = { "hello" }
+   ]])
+
+   it("child fields do not exist in parent", util.check_type_error([[
+      local Node = record
+         tag kind: string
+         x: number
+         y: number
+      end
+
+      local EnumNode = record is Node with kind = "enum"
+         enumset: {string}
+      end
+
+      local e: Node = {}
+      e.x = 12
+      e.y = 13
+      e.enumset = { "hello" }
+   ]], {
+      { msg = "invalid key 'enumset' in record 'e' of type Node" }
+   }))
 end)
--- a/spec/assignment/to_nominal_record_spec.lua
+++ b/spec/assignment/to_nominal_record_spec.lua
@@ -1,4 +1,5 @@
 local tl = require("tl")
+local util = require("spec.util")
 
 describe("assignment to nominal record", function()
    it("accepts empty table", function()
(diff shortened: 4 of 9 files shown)


luauser32167:
What do you think about a more concise declaration syntax (similar to Rust's enums)?:

```lua
local MyTagRec = taggedrecord
    Foo = record
        a: number
    end

    Bar = record
        a: string
        b: boolean
    end

    Baz = record
        a: Point
        b: Error
        c: any
    end
end

local tr1 = MyTagRec.Foo({ a = 1.0 })
local tr2 = MyTagRec.Bar({ a = 'a', b = false })
local tr3 = MyTagRec.Baz({ a = {}, b = {}, c = nil })
local trs: {MyTagRec} = { tr1, tr2, tr3 }

-- raw tag access
--
for _, tr in ipairs(trs) do
    local tag: string = tagof(tr)
    if tag == 'Foo' then
        ...
    elseif tag == 'Bar' then
        ...
    elseif tag == 'Baz' then
        ...
    end
end

-- exhaustive case analysis
--
for _, tr in ipairs(trs) do
    switch tr

    when Foo do
        ...
    end

    when Bar do
        ...
    end

    -- error missing 'Bar'
    -- when Bar do
    --     ...
    -- end

    end -- switch
end

-- non exhaustive case analysis
--
for _, tr in ipairs(trs) do
    if tr is Foo then
        ...
    else
        ...
    end
end
```

pdesaulniers:
I'm not speaking for @hishamhm, but:

I think the current syntax is convenient for interfacing with existing Lua libraries which use some form of 'tagged table' pattern. [Here's a declaration file which illustrates this.](https://github.com/teal-language/teal-types/blob/5685206e41b0b374a3c693da3b3ff02dd87fdac1/types/SLAXML/slaxdom.d.tl#L1-L21)

Existing libraries may use any key name for the tag. For instance, in @hishamhm's  example above, the tag is called `widget_type`. SLAXML calls it `type`. I think the syntax needs to support this.

euclidianAce:
I have some concerns about emulating some OOP with this
 - Multiple inheritance seems clunky, having to tag both the super and sub class with differently named tags. For example in my lmenu module that I would like to translate to Teal, I have a class structure that I think would translate to:
```
local Menu = record
   tag mType: string
   options: {Option}
end

local List = record is Menu with mType = "list"
   tag mSubType: string
  -- [[ ... ]]
end

local Checklist = record is List with mSubType = "checklist"
  -- [[ ... ]]
end
```
This is definitely manageable, but inconvenient.
 - What if we want to extend a module's record, but it isn't tagged, what should be done? The only workaround that I can think of is just creating a new record with the modules record as an entry.
```
local foo = require("foo")

local myType = record
   tag mType: string
   _foo: foo.bar

end

local mySubType = record is myType with mType = "blah"
   baz: number
end
```
 - Also, how are generics handled?
```
local myType = record<T>
   tag mType: string
   data: T
end

-- is the generic implied?
local mySubType1 = record is MyType with mType = "a"
   extraData: T
end

-- do we need to specify twice?
local mySubType2 = record<T> is MyType<T> with mType = "b"
   extraData: T
end
-- This would be my guess as to allow for
local mySubType2_1 = record<T, K> is MyType<T> with mType = "c"
   extraData: K
end

-- does this work?
local mySubType3 = record is MyType<number> with mType = "d"
   extraData: number
end
```

hishamhm:
@euclidianAce that it good feedback. This model indeed does not cover multiple-level inheritance well and shouldn't be seen as a "clunky way of doing OOP". I think for OOP we actually need a `class` model.

> Also, how are generics handled?

There are most likely holes in there, I didn't consider generic tagged records at all when writing this. I wanted to get it out quickly as a proof-of-concept to get feedback. Thanks for raising these points.

I still have reservations with the ideas in this PR myself, it's one of the reasons why I didn't merge.

fgaz:
I think this could benefit from allowing a bit more expressiveness.
Sometimes, the type can't be discriminated by looking at the value of a field. Maybe you have to look at the _presence_ of a field. Maybe the type is opaque and you have to call a library function.

I propose that `with` is to be followed by an expression of type `function(parent_record): boolean`. The function will be called wherever a corresponding `is` expression is found.

The example in OP would translate to

```lua
local WidgetType = enum
   "button"
   "field"
end

local Widget = record
   widget_type: string -- no longer a special "tag"

   x: number
   y: number
end

local Button = record is Widget with function(w) return w.widget_type == "button" end
   text: string
end

-- Example of separate isX definition
local function isField(w)
  return w.widget_type == "field"
end
local Field = record is Widget with isField
   width: number
   label: string
   input: string
end

local function draw(x: number, y: number, w: number, t: string)
end

local function show_widget(w: Widget)
   if w is Button then -- this generates: if (function(w) return w.widget_type == "button" end)(w) then
      draw(w.x, w.y, 10, w.text)
   elseif w is Field then -- this generates: elseif (isField)(w) then
      draw(w.x, w.y, w.width, w.label)
      draw(w.x, w.y, w.width, w.input)
   else
      draw(w.x, w.y, 10)
   end
end

local widgets: {Widget} = {
   { widget_type = "field", width = 20, label = "Name:", input = "" } as Widget,
   { widget_type = "field", width = 3, label = "Age:", input = "" } as Widget,
   { widget_type = "button", text = "OK" } as Widget,
   { widget_type = "button", text = "Cancel" } as Widget,
}
```

mypy and typescript have similar features:

* https://mypy.readthedocs.io/en/stable/type_narrowing.html#user-defined-type-guards
* https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

---

Problem 2: OP is basically inheritance, but what if `Button` is both a `Widget` and something else? Are multiple `is ... with` expressions allowed? What if `Button` is defined in a library and I need to use it in a union in my code? I can't add `is ... with` to the library code, and if I could teal would lose exhaustivity checking.

On the other hand, making it so the union is defined in the top-level record (tagged union / sum type) prevents the addition of "subclasses".

This is basically the [expression problem](https://en.wikipedia.org/wiki/Expression_problem).

In the case of teal, I think tagged unions / sum types would be more useful. I feel like grouping up existing types in different ways comes up more often, but maybe I'm just ignorant of the oop side of lua

Frityet:
How usable/ready is this PR? Being able to discriminate between multiple record types is the big thing I really need in a codebase I am working on right now, so I am considering using the version of Teal in this PR

hishamhm:
Closing this PR since I have implemented `where` annotations for discriminating `record` and the new `interface` types in the `next` branch, which allows for discriminating records in unions!

hishamhm:
> Existing libraries may use any key name for the tag. For instance, in @hishamhm's example above, the tag is called widget_type, whereas SLAXML calls it type. I think the syntax needs to support this.

@pdesaulniers the new syntax in the `next` branch is `record Foo is Bar where self.x == "y"`, which allows for an arbitrary expression using `self`. This should cover the use of different tag fields of different types and even things like method calls for checking types (e.g. `where self:typeOf("Drawable")` [as used in Love2D](https://www.love2d.org/wiki/Object:typeOf) — though I haven't tried updating Love2D type definitions myself!)

hishamhm:
> Multiple inheritance seems clunky

@euclidianAce You can merge multiple interfaces with `is` (e.g. `record R is I1, I2, I3 where ...`) but your `where` expression is up to the programmer. Each interface can have its own `where` clause, so you can do `if my_r is I1` and that will translate to I1's `where` expression, but again, it's up to the programmer to ensure that those tests are consistent with the object model. Given the multiplicity of Lua object models, I feel this is the best we can do, save for introducing yet another one (which then wouldn't interop...).

> What if we want to extend a module's record, but it isn't tagged

In `tl.tl` I have used tricks such as `interface RecordLikeType is Type where self.fields` abusing the presence of a known fields to mean that the table is that type. If you have control over the various shapes of data in your program, it's often possible to detect them. Worst case, one can write a more complex checker function and use that instead: `interface Foo where am_i_foo(self)`

> Also, how are generics handled?

I haven't sorted this out yet, even in the `next` branch. I wanted to get _something_ done and figured I could handle the plain case first and tackle the generic case later.

Our generics are still pretty limited anyway (but already super useful), so I figured that even an "incomplete" feature would already be a step in the right direction rather than trying to figure everything out ahead of time.

hishamhm:
> I propose that with is to be followed by an expression of type function(parent_record): boolean. The function will be called wherever a corresponding is expression is found.

@fgaz I ended up implementing something like this in the `next` branch! But it uses macro-expressions to allow for efficient checks in the common case. You write `with <expression>` where the `<expression>` is allowed to use the `self` identifier once, which allows for a simple in-place replacement over `if v is T` (If you need to use it multiple times, you can always do `where f(self)` which then makes `if v is T` translate to `if f(v)`)

> Problem 2: OP is basically inheritance, but what if Button is both a Widget and something else? Are multiple is ... with expressions allowed? 

Records and interfaces in the `next` branch accept multiple `is <interfaces...>` entries (not multiple records), and each interface can have its own `with` test at its own definition (see the `next` branch's `tl.tl` for examples).

PR #128, opened 2020-06-03
Sort alphabetically the 'standard_library' talbe's fields

luauser32167:
When the fields are sorted it's easier to see which functions have been annotated with types and which are not yet annotated.


hishamhm:
@luauser32167 Thanks for the PR! Does this PR add or change any of the definitions or it just reorders them?

luauser32167:
It only sorts them. By doing so I think it becomes easier to find the place where the other missing type definitions can be added.

luauser32167:
PR #130 has the _standard_library_ table's fields sorted.

PR #132, opened 2020-06-04
Add primitive thread type and coroutine library

3uclidian:


hishamhm:
This is great, thank you! 

By its very nature, the `coroutine` API is very dynamic so I figure that using coroutines with Teal will involve a lot of casts, but I guess that's the nature of the thing.

I'm considering exposing a second-class tuple type (they already exist internally) that could be used in casts to improve the experience with multiple returns, but don't know how much that would really help things (adding wonky second-class types is not something to be done lightly!)

PR #139, opened 2020-06-08
tl: report when there is an error loading tlconfig.lua

euclidianAce:
tl doesn't report syntax errors in tlconfig making it hard to track type errors from .t.dl files that are actually due to a missing comma or something in tlconfig.lua.


hishamhm:
That's a nice improvement! Thanks!

PR #142, opened 2020-06-08
add xpcall function

cloudfreexiao:


tl.tl
```diff
@@ -3370,6 +3370,7 @@ local standard_library: {string:Type} = {
       a_type { typename = "function", args = {}, rets = { ALPHA, BETA } },
    } },
    ["pcall"] = a_type { typename = "function", args = { VARARG_ANY }, rets = { BOOLEAN, ANY } },
+   ["xpcall"] = a_type { typename = "function", args = { VARARG_ANY }, rets = { BOOLEAN, ANY } },
```
hishamhm, line 3373: on second thought, this should probably be

`args = { FUNCTION, FUNCTION, VARARG_ANY }`

and pcall`s should be `args = { FUNCTION, VARARG_ANY }` :thinking:

cloudfreexiao: i think so :)

PR #143, opened 2020-06-08
Add more details on use of .d.tl files

factubsio:


docs/tutorial.md
```diff
@@ -737,3 +731,90 @@ Note that even though adding type annotations to .lua files makes it invalid
 Lua, you can still do so and load them from the Lua VM once the Teal package
 loader is installed by calling tl.loader().
 
+
+### Type definitions for third party libraries
+
+You can also create declaration files to annotate the types of third-party Lua
+modules, including C Lua modules. To do so, create a file with the .d.tl
+extension and require it as normal, i.e. `require("minetest")`.
```
hishamhm, line 739: It would probably be better to pick some non-global module as a first example, such as `local lfs = require("lfs")`.

docs/tutorial.md
```diff
@@ -737,3 +731,90 @@ Note that even though adding type annotations to .lua files makes it invalid
...
+
+Types defined in this module will will be used as a source of type information 
+checking with `tl check`, even though the real Lua module will be loaded
+instead when requiring the module from Lua or `tl run`.
+
+If the third party library is "globally available", i.e. you do not explicitly
```
hishamhm, line 745:
```suggestion
If the third party library is "globally available" in your execution environment, i.e., if you do not explicitly
```

docs/tutorial.md
```diff
@@ -737,3 +731,90 @@ Note that even though adding type annotations to .lua files makes it invalid
...
+end
+
+return {
+    MyPointType = MyPointType,
+    MyCompositeType = MyCompositeType,
+}
+```
+
+will give errors when accessing `MyCompositeType::center`, as `MyPointType` is not
+visible.
```
hishamhm, line 781: ^ Are you sure you're running the latest tl? This section is not correct as of tl 0.6.0 and `master`! The good news is that above example actually works, you can use it like this:

```
local mypoint = require("mypoint")

local t: mypoint.MyCompositeType = {}

t.center.x = 2

local p: mypoint.MyPointType = { x = 2, y = 3 }

t.center = p
```

docs/tutorial.md
```diff
@@ -737,3 +731,90 @@ Note that even though adding type annotations to .lua files makes it invalid
...
+return MyCompositeType
+```
+
+This will mean that references to `MyPointType` must be qualified (or locally declared) as
+`MyCompositeType.MyPointType`.
+
+##### Global Types
+
+While perhaps discouraged in general, for types that are provided as part of the
+execution environment it may be more hygeinic to declare them as global.
```
hishamhm, line 807: Appreciate the care you took here, but this is a little tricky to give as advice, because the technically the execution environment does not provide these types for real (at least not the type _names_) so ocuppying global names with type definitions is not something to be done lightly. 

Given that exported local and indirectly referenced types work, as described above, I'm inclined we should avoid directing users to global types, at least in the tutorial.

Perhaps instead of this subsection, we could have just a quick note that "Types can be declared with `global`, but in general this should be avoided in reusable modules."

docs/tutorial.md
```diff
@@ -737,3 +731,90 @@ Note that even though adding type annotations to .lua files makes it invalid
...
+checking with `tl check`, even though the real Lua module will be loaded
+instead when requiring the module from Lua or `tl run`.
+
+If the third party library is "globally available", i.e. you do not explicitly
+`require` it, you can inject the types without affecting your generated code
+with a `tlconfig.lua`:
+
+```lua
+return {
+    preload_modules = {
```
hishamhm, line 751: (nitpick: all code examples in this doc are indented with 3 spaces)

hishamhm:
@factubsio Left a bunch of review comments! Many many thanks for the initiative!!

docs/tutorial.md
factubsio: Good call, I just grabbed the one out the issue that discussed it!

docs/tutorial.md
factubsio: If they are injected via preload they shouldn't end up in the generated code at all (unless I am missing something)?

Aye if indirectly referenced types work (I will drag in latest tl tomoz) then 100% agreed that's a nicer way to do it.

docs/tutorial.md
factubsio: As in - the code examples *should* be indented with 3 spaces?

docs/tutorial.md
factubsio: I originally had this, not sure why I changed it, but if the preference is this way then 👍 from me.

docs/tutorial.md
factubsio: Huh, moves fast, great. If I don't have to re-do argparse and compat53 I should be able to update pretty quickly. (luarocks on windows with vs2019 is... unpleasant).

docs/tutorial.md
hishamhm: yup, to keep visual consistency with the ones that are already there

factubsio:
> @factubsio Left a bunch of review comments! Many many thanks for the initiative!!

Will do a big old change once I manage to get a newer tl going locally. Thanks for the quick reply.

docs/tutorial.md
factubsio: Gotcha, I thought I had copy/pasted wrong when I saw 3-space indents. 😬

docs/tutorial.md
hishamhm: yeah, not in the generated code, but in the symbol table during compilation. Say I have my own type Bla in my application, and I want to use library foo, and it happens to expose its own global type Bla. Then my own type Bla is shadowing Bla from module foo. If module foo exposed it as foo.Bla, then I could use both with no problems

factubsio: Gotcha. All makes sense now.

hishamhm:
@factubsio Ping :) Just checking in if you're still interested in reviewing this! Updating your local Teal should be as easy as grabbing the latest `tl.lua` file and `tl` main script from the repository and replacing them. 

Also, since you mentioned you're running on Windows: I have recently built a standalone `tl.exe` binary (which bundles Teal, a Lua interpreter and all its dependencies) — would you be interested in that?

docs/tutorial.md
```diff
@@ -737,3 +731,90 @@ Note that even though adding type annotations to .lua files makes it invalid
...
+#### Visibility
+
+It should be noted that local references in type definition files (or any files
+containing types for that matter) are not available in the requiring module.
+
+This means that constructs like:
+
+```lua
+local MyPointType = record
+    x :number
```
bew, line 766: Could you please keep the consistency of where the `:` is placed?
In the rest of the file, the colon is before the space, like `var: type` and in most of the code you added you place it after the space, like `var :type`.

I think it should be consistent in the same file, so change the added code to use `var: type`

pdesaulniers:
This has been merged as https://github.com/teal-language/tl/commit/f22e0f5a7a89c393ba60c804ddb6ddb4f70a8bcc.

Thanks!

PR #147, opened 2020-06-09
Add --version to tl

factubsio:
There might be a nicer way of keeping the version_string up to date? (I guess it's only used in CHANGELOG)

luarocks knows the version too, but I have no idea how that is kept in sync.

### DIFF
--- a/tl
+++ b/tl
@@ -1,5 +1,7 @@
 #!/usr/bin/env lua
 
+local version_string = '0.7.0'
+
 local function script_path()
    local str = debug.getinfo(2, "S").source:sub(2)
    return str:match("(.*[/\\])") or "."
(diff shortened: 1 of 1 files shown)


tl
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env lua
 
+local version_string = '0.7.0'
```
hishamhm, line 3:
```suggestion
local version_string = "dev"
```

Let's keep it marked as "dev" in the Git repo so that it's easier to identify when people are running code downloaded straight from the repository as opposed to releases from LuaRocks.

I'll need to set something up on my end to make sure I update this file whenever I upload a new release to LuaRocks.

I've looked at this, but it doesn't look like I can make it fully automatic whenever users download it from Git:
https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes#_keyword_expansion

factubsio: Would you mind if it was "0.7.0-dev" (or best "0.7.0-short_sha") - should make it even easier to to tell what people are running from git.

I get confused pretty consistently which version I'm currently on, even working from git.

hishamhm: > "0.7.0-short_sha"

that has a chicken-and-egg problem because the sha for a commit depends on its contents (see link above). But yeah, I guess "0.7.0+dev" could work!

Issue #137, opened 2020-06-06
`typealias` construct?


pdesaulniers:
Teal supports `functiontype` to avoid longwinded function type declarations. In that case, why not add a generic construct for creating type aliases of any kind? This could be useful with union types:
```lua
local Something = typealias(number | string | SuperLongTypeName)
local x: Something = 0
```

hishamhm:
That's a good idea, and if it accepts a general `type` (as in the grammar.md rule `type`), then it even renders `functiontype` redundant since it could be instead a `typealias(function() ...)` instead. I wouldn't mind replacing it with something more general, never liked the name `functiontype` a lot.

Not sure I like the parentheses in the syntax because it makes it look like a function call, but since types can be parenthesized at the top level, I guess even without them something like `typealias (number | string | SuperLongTypeName)` would be valid anyway.

pdesaulniers:
I'm satisfied with https://github.com/teal-language/tl/pull/194. I think https://github.com/teal-language/tl/pull/159 and
https://github.com/teal-language/tl/issues/170 can be closed as well.

PR #159, opened 2020-06-11
Introduce typealias, remove functiontype

hishamhm:
Introduce `typealias` as suggested in #137, remove `functiontype` because it becomes redundant with `typealias function`.


pdesaulniers:
It seems to be working great!

Food for thought: should the type aliases be expanded in the error messages? Here is the current behavior:

test.tl:
```lua
local MyWeirdType = typealias number | string | {string}
local x: MyWeirdType = {1, 2, 3}
```
`tl check` result:
```
1 error:
test.tl:2:24: in local declaration: x: got {number}, expected MyWeirdType
```

Instead of displaying the aliased type (`expected MyWeirdType`), maybe we could display the real type:
```
test.tl:2:24: in local declaration: x: got '{number}', expected 'number | string | {string}'
```

hishamhm:
@pdesaulniers I sort of considered this, but it's a bit tricky to pull off, because in the current implementation, typealiases are really nominal types that represent their underlying structure. So if I expand those names on errors, the same would happen with named record types as a side-effect `got number, expected {x: number, y: number}` instead of `got number, expected Point`, and that would make it look as if the type system was structural (which isn't true, it is nominal).

hishamhm:
Closing this old PR since a different version of this was implemented by adding the `local type` syntax!

PR #161, opened 2020-06-11
Mention text editor support in the README

pdesaulniers:
Thanks to @euclidianAce, Teal language support for Vim is now quite good! I think we can advertise the plugin in the README.

I've also added a note about vscode support.

@mrtnpwn, you said you needed to update https://github.com/mrtnpwn/tl-nano?


hishamhm:
Good call! Merged!

PR #164, opened 2020-06-12
Add -q --quiet flag

pdesaulniers:
This flag suppresses information messages on stdout.


hishamhm:
Would -q,--quiet be a more "standard" name for this or is there some semantic difference?

pdesaulniers:
I have no idea :) How is this kind of flag named in other tools? At least, I know `make` supports both `--silent` and `--quiet`...

pdesaulniers:
Yeah, actually, `-q --quiet` might be better, since `--silent` sounds like it should also silence warnings and errors.

PR #174, opened 2020-06-15
feat(lib): add gen, process_string methods

### LINKED ISSUE #171: add generated output as available tl lib method
It appears that tl can only process files, via `process`. It would also be nice if the cli commands of `gen` and `run` were available at the library level. I created a new method `tl.process_str`, by refactoring out the file io of `process`. Let me know if this sounds good, open to naming and function signatures. Right now it's:

```lua
tl.process(filename, env, result, preload_modules)
-- new method:
tl.process_str(input, env, result, preload_modules, is_lua, filename)
```

```lua
-- playground.lua
local tl = require('tl')

local code = [[
local message:string = "hello"
]]

local result = tl.process_str(code)
print(tl.pretty_print_ast(result.ast))
```

Maybe there be a function `tl.gen(input)` that returns the pretty printed string for you?

Background: this is helpful for the teal browser "playground"

darrenjennings:
Adding `tl.gen` and `tl.process_string` so that `tl`'s string processing will not be limited to file io. This will be helpful in building the browser playground, and maybe more use cases?

I noticed that my local LUA_PATH did not have `./?.lua` so that tl tests were always running tests against the source from the registry. I've added a busted config so that `lpath` can use non-registry/local `tl`.

Fixes #171

### DIFF
--- a/.busted
+++ b/.busted
@@ -1,0 +1,5 @@
+return {
+  default = {
+    lpath = "./?.lua";
+  }
+}
--- a/spec/api/gen_spec.lua
+++ b/spec/api/gen_spec.lua
@@ -1,0 +1,43 @@
+local tl = require("tl")
+local util = require("spec.util")
+
+describe("tl.gen", function()
+   it("can process tl strings", function()
+      local input = [[
+         local movie:string = "Star Wars: Episode"
+         local episode: number = 4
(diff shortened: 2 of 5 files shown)


hishamhm:
woohoo, merged! :tada:

Issue #67, opened 2020-03-10
Add some way to bundle multiple .tl scripts as one file?


pdesaulniers:
I think it would be nice if we could use a command-line argument to bundle multiple tl scripts as a single Lua file. TypeScript supports this by using the `--outFile` argument.

Right now, tl's source code is contained inside a single .tl file. If this feature gets implemented, then tl's source code could be split into multiple files. This would make the code a bit easier to browse.

hishamhm:
> I think it would be nice if we could use a command-line argument to bundle multiple tl scripts as a single Lua file. TypeScript supports this by using the --outFile argument.

That would be a cool feature, but would probably accelerate to the desire of some sort of source maps, because the line numbers in the tracebacks when running tl code would not match the input anymore (so this is a feature request that triggers another feature request :) ). Still, a patch would be welcome!

> If this feature gets implemented, then tl's source code could be split into multiple files. This would probably make the code a bit easier to browse.

I'm still a bit on the fence for that tbh, but it's a thought to consider for sure!

ash-nolan:
Just my 2¢: I think bundling should be handled by a separate utility.
Mapping `.tl` files one-to-one onto `.lua` files fits the textbook definition of a compiler and seems to align well with the Unix philosophy.

Bundling doesn't particularly sound like something that needs to be integrated into the compiler, or really even something that needs to be Teal-exclusive.
I think the use case is valid, but perhaps that use case could be addressed by a separate utility under the `teal-language` family: that way the compiler can be kept minimal.

jirutka:
This can be achieved using [luapak](https://github.com/jirutka/luapak) merge.

hishamhm:
@jirutka would love to see luapak eventually convert .tl files to .lua in that same step! :) Do you think `os.execute` to run `tl gen` would suffice or would a stable programmatic API would make more sense?

jirutka:
A stable programmatic API would be better. :)

siffiejoe:
I have recently added 3rd-party plugin support to a [tool of mine][1] that can bundle multiple Lua modules into one Lua file. Apparently, those plugins can also be used to transpile files in a Lua dialect on-the-fly during amalgamation. As a proof-of-concept, I have added plugins for moonscript, fennel, and [teal][2].

AFAICS, a similar approach could be used with luapak -- maybe even with the same plugins.

[...]

Issue #47, opened 2020-02-05
Out of source builds?


pdesaulniers:
I think it would be useful if the CLI supported some way to generate Lua scripts into a separate directory. For instance,
```
tl --outdir=build/ gen game/main.tl
```
...would generate
```
build/game/main.lua
```
That way, I would not have to pollute my source tree with .lua files. I could simply add the `build` directory to `.gitignore`.

hishamhm:
+1 to this feature!

euclidianAce:
@hishamhm i believe #177 closes this

pdesaulniers:
Yep!

Issue #107, opened 2020-05-13
Common data structure for CLI and project file


iljio:
Hello,

I think there should be a common data structure that can be referenced by the parsers for the CLI and for the project file, and I think the parsers should create a keyvalue structure that gets sent to the compiler. The compiler should then decide how to make use of the data.

I think commands can be represented as tables within this common data structure table, where each command defines their own aliases (to use with matching user input), descriptions (for the help text), and list of parameters. Each parameter could implemented as a table that defines its own aliases, description, optional requirement, default value, and intended Lua type (for type-checking and conversion of user input).

We could restrict flags so they represent keyvalues, i.e. `<key>=<value>` You could have a key like `target_version` that only takes 5.1, 5.2, 5.3, or LuaJIT2 as an applicable value, and the relationship between the CLI flag and the data that gets sent to the compiler is 1:1 (ignoring any parsing errors of course!). Thus, flags would get their own table, where each flag defines its own aliases, description, default value (for when the user doesn't specify the flag), and intended Lua type.

All together, this can enrich the help text/documentation as well, as all of the possible data that exists ends up well-defined in the form of a Lua table. Coercing that information into different formats would be easy, and I would imagine it lays the groundwork for project file implementation.

Plus, extending the CLI/project file is a simple as modifying the data structure and implementing use of the additional or modified data in the compiler itself.

I would love to program this, but as I'm a community project newbie, I wanted to start a dialogue about it.

Issue #31, opened 2020-01-29
Convenient way of generating Lua files for every tl file?


pdesaulniers:
Right now, it seems like I need to manually call `tl gen <filename>` for every tl file in my project.

While I can write a script to automate this, I suppose it would be nice if there was a built-in way of generating Lua files for multiple tl modules at once.

For instance, in TypeScript, we can specify the input files inside [tsconfig.json](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html) and run `tsc` to compile the whole project.

mrtnpwn:
You can use [Tup](http://gittup.org/tup) for that I guess

```tup
: foreach *.tl |> tl gen %f |> %B.lua
```

I haven't test it yet but that's how it works.

hishamhm:
I guess we can call this one fixed by #42, since you can run `tl gen *.tl` now!

pdesaulniers:
Yes, that works. However, I would prefer if there was some way to specify the input files inside `tlconfig.lua`. That way, users would not have to rely on a shell-specific script or command to build the project.

For instance, the syntax for globs is shell dependent, especially when we get to patterns like `**/*.tl`. Also, my `build.sh` script would not work in cmd.exe :(

I'd be happy if tlconfig supported something akin to the ["include" and "exclude" properties in tsconfig.json](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#examples). This would allow using globs in a platform independent manner. 

For the time being, though, I suppose a simple `files` array would be good enough.

hishamhm:
> For the time being, though, I suppose a simple files array would be good enough.

Yeah, that should be an easy add. Anything involving globs would take some extra dependency to do the directory traversal.

pdesaulniers:
It seems like this library could do the job: https://github.com/starwing/lpath. It seems a tiny bit obscure, but it does support globs:
```lua
local fs = require("path.fs")

-- find all .lua files in directory "game"
local src = "game/*/*.lua"

local files = fs.glob(src)

for _, v in ipairs(files) do
	-- do something
end
```
This library could also be used for building paths in https://github.com/hishamhm/tl/issues/47.

Are you aware of a better library for this?

hishamhm:
> Are you aware of a better library for this?

[...]

PR #177, opened 2020-06-23
tl: add build options

euclidianAce:
Adds luafilesystem dependency and the following config options
- build_dir
- source_dir
- files
- include
- exclude
- Original include -> include_dir

Hopefully these changes should remove the need for a Makefile for pure Teal projects (though probably not `tl` itself)

This contains some pretty big refactors to `tl` that I will try to outline here.
- Setting up the environment and preloaded modules has been abstracted into some methods for reusability
- `tl gen` and `tl check` now build up a source map for file names (`src_map: {string:string}`)
	- `tl run` will not build up a source map because it only relies on the args passed to it.
	- In the future there could probably be a `source_map` config option as well to just directly map inputs and outputs.
- There is some custom logic to recursively traverse sub-directories and build up the source map, This could probably be optimized in the future.

Some notes:

- Technically the lfs dependency can be made optional and only required to use things like `source_dir`. Technically `build_dir` can get by with some `os.execute("mkdir ...")` stuff, but that always feels hacky.
- This currently passes the test suite but I'm am unsure of how to write tests for these features, so help on that would be appreciated.
	- Especially with the `**/` pattern matching functions, there doesn't seem to be a way to extract local functions with the debug library, so putting them in a module seems like the only way to test them, but having a test suite for these would be helpful.

Constructive criticism/reviews are definitely appreciated, especially for testing this. :tada:


euclidianAce:
(Should close #47, #31, and maybe #107?)

tl-dev-1.rockspec
```diff
@@ -16,6 +16,9 @@ dependencies = {
 
    -- needed for the cli tool
    "argparse"
+
+   -- needed for --source-dir
+   "lfs"
```
hishamhm, line 21: the dependency name for LuaRocks is called `luafilesystem` (I know...)

hishamhm:
@pdesaulniers Could you take a look at this PR? :)

pdesaulniers:
OK, I'm looking at this.

tl-dev-1.rockspec
```diff
@@ -16,6 +16,9 @@ dependencies = {
 
    -- needed for the cli tool
    "argparse"
```
pdesaulniers, line 18: There's a missing comma here (@hishamhm, I think CI should've caught this?)

pdesaulniers:
> Especially with the **/ pattern matching functions, there doesn't seem to be a way to extract local functions with the debug library, so putting them in a module seems like the only way to test them, but having a test suite for these would be helpful.

So far, the CLI flags are tested by calling `tl` from the shell and looking at its output, like this: https://github.com/teal-language/tl/blob/5732126a259a67908d7017b3a150c00d5a97bfd5/spec/cli/include_spec.lua#L4-L16

Right now, it would be a bit difficult to write this kind of tests for `include` and `exclude`, because  `include` cannot be set with a command line option, and `tl` assumes that `tlconfig.lua` is contained inside the working directory.

So, for now, I would suggest to add some command line flags for `include` and `exclude`, and use these in the tests.

Otherwise, we could also add a `-p --project` flag for setting the path of `tlconfig.lua`, but this sounds a bit harder to manage (we might need one `tlconfig.lua` per test)

pdesaulniers:
> (Should close #47, #31, and maybe #107?)

Yes for the first two, but I'm not sure about https://github.com/teal-language/tl/issues/107. I think that issue is more about merging all of the command line and `tlconfig.lua` stuff inside a single table. That way, I suppose we could get rid of the [valid_keys array](https://github.com/teal-language/tl/blob/5732126a259a67908d7017b3a150c00d5a97bfd5/tl#L39-L43), generate the [list of compiler options](https://github.com/teal-language/tl/blob/master/docs/compiler_options.md#list-of-compiler-options) programmatically, etc.

euclidianAce:
> tl assumes that tlconfig.lua is contained inside the current working directory.

Since we're bringing in lfs for this feature, we could probably use `lfs.chdir` and build up some directories in /tmp (maybe copying the tl executable too?) and just throw some `helloworld.tl`s and the `tlconfig.lua`s that we want to test in there. When I get the chance I'll try to write something up.

tl-dev-1.rockspec
hishamhm: oh good catch, that's right

Pushing #178 to catch this

hishamhm:
@euclidianAce sounds like a plan!

docs/compiler_options.md
```diff
@@ -24,5 +24,81 @@ return {
 | Command line option | Config key | Type | Description |
 | --- | --- | --- | --- |
 | `-l --preload` | `preload_modules` | `{string}` | Execute the equivalent of `require('modulename')` before executing the tl script(s). |
-| `-I --include` | `include` | `{string}` | Prepend this directory to the module search path.
+| `-I --include-dir` | `include_dir` | `{string}` | Prepend this directory to the module search path.
 | `--skip-compat53` | | | Skip compat53 insertions.
+|| `include` | `{string}` | The set of files to compile/check. See below for details on patterns.
+|| `exclude` | `{string}` | The set of files to exclude. See below for details on patterns.
+| `-s --source-dir` | `source_dir` | `string` | Set the directory to be searched for files. `gen` will compile every .tl file in every subdirectory by default.
+| `-b --build-dir` | `build_dir` | `string` | Set the directory for generated files, mimicing the file structure of the source files.
```
pdesaulniers, line 32: Minor nitpick: should be 'mimicking' instead of 'mimicing'

docs/compiler_options.md
```diff
@@ -24,5 +24,81 @@ return {
...
+|| `files` | `{string}` | The names of files to be compiled. Does not accept patterns like `include`.
+
+### Include/Exclude patterns
+
+The `include` and `exclude` fields can have glob-like patterns in them:
+- `*`: Matches any number of characters (excluding directory separators)
+- `**/`: Matches any number subdirectories
+
+In addition, setting the `source_dir` has the effect of prepending `source_dir` to all patterns
```
pdesaulniers, line 41: Minor nitpick: I would add a period at the end of this sentence.

euclidianAce:
I've made some pretty big refactors to this now that I've been able to test it and I want to just outline them here:

- Previously `tl gen` was modified to build a project when run with no arguments. This has been reverted back and I've added `tl build` which has that behavior
- `tl build` has a new flag associated with it (inspired from Gentoo)
    - `-p --pretend`: don't compile, but type check and print what files would be written to
        - Similarly from Gentoo, an `--ask` flag could easily be implemented as well
- The current `tl gen` now works more like gcc:
    - files are generated in the current directory unless specified with `-o --output`
    - multiple files are concatenated into one when using -o
        - This feature practically comes for free, by just appending instead of writing. With how the current compiler works, source maps may not be all that hard to implement since there is no minifying.
        - Idk it that's enough to close #67 but its a start.
- `--skip-compat53` can be put in `tlconfig.lua` as `skip_compat53 = true`, unrelated, but I'm using this pr with some of my own projects and it was annoying to type each time.

The test coverage is pretty alright, but I still need to write more :P

hishamhm:
@euclidianAce wow, impressive!!

Some commentary based on the above comment (haven't looked deeply into details of the code yet)

> Previously tl gen was modified to build a project when run with no arguments. This has been reverted back and I've added tl build which has that behavior

Nice!

> -p --pretend: don't compile, but type check and print what files would be written to 

I've seen this being called `--dry-run` in other programs, but I guess `--pretend` works just as well!

> multiple files are concatenated into one when using -o 

This bit I am a little concerned with. If multiple files just declare global functions, then this will work fine, but if each file declares a module, then mere concatenation of the outputs will not produce what I would consider the expected behavior: amalgamation scripts generally end up declaring multiple independent modules (like one would see on a big `.so` C module that declares multiple Lua modules). I would rather drop this feature for `tl gen` for now (producing an amalgamation script could be done as a separate feature in a separate PR — let's restrict PR scope to get this one merged sooner :) )

> --skip-compat53 can be put in tlconfig.lua as skip_compat53 = true, unrelated, but I'm using this pr with some of my own projects and it was annoying to type each time.

:+1: 

> The test coverage is pretty alright, but I still need to write more :P

If the coverage is alright I guess we can look into reviewing and merging this version. Let's just make sure the new features are covered in the docs.

Very cool work in this PR, improving the tooling is super important!

euclidianAce:
For the time being I'll limit `--output` to asserting only one script argument. Didn't really think of the consequences of that one.

But this is pretty feature-complete for what I imagined the initial write up of this would be and while I still have some ideas, I think it would be feature creep. So over the next few days I'll try to improve test coverage and documentation to get this ready to merge. :tada:

euclidianAce:
I'll leave this here as a reference for where I think I'm at and suggestions for where to go.
## TODO

- Test:
	- [x] `source_dir`
	- [x] `build_dir`
	- [ ] any file including option with relative vs non-relative paths
	- [x] `include+exclude` globbing
	- [x] basic interactions between `include+exclude`
	- [ ] `include+exclude` with multiple patterns
	- [x] basic interactions between `source_dir+build_dir`
	- [x] basic interactions between `source_dir+build_dir+include+exclude`
	- [ ] error handling for `source_dir`
		- i.e. `source_dir` not found: error out gracefully, probably mentioning source dir in some way
	- [ ] error handling for `build_dir`
		- there is some error handling, but not extensively tested yet
	- [x] `-o --output`
	- [ ] `-p --pretend --dry-run`
		- honestly this could just be a copy-paste of any instance where `util.run_mock_project` was used and just add `-p` and verify the output
	- [ ] `tl build`
		- any usage of `util.run_mock_project` is basically the test suite for `build`, but there could be more explicit tests
- Document:
	- [x] `-s --source_dir`
	- [x] `-b --build_dir`
	- [x] `files` tlconfig.lua entry
	- [x] `include+exclude`
		- [x] globbing patterns
	- [x] `-o --output`
	- [x] `-p --pretend --dry-run`
	- [x] `tl build`
	- [x] `skip_compat53` tlconfig.lua entry

In general, I think I need to write more tests where the expected output is an error.

tl
```diff
@@ -245,92 +300,456 @@ for i, filename in ipairs(args["script"]) do
...
+      if not s_idx then
+         matches = false
+         break
+      end
+   end
+   return matches
+end
+local function matcher(str)
+   local chunks = {}
+   for piece in str_split(str, "**/") do
```
hishamhm, line 511: since the rest of the code is using `path_separator`, should this use as well?

tl
```diff
@@ -245,92 +300,456 @@ for i, filename in ipairs(args["script"]) do
...
+   local current_dir = self.dir
+   for dirname in str_split(path, path_separator) do
+      current_dir = current_dir[dirname]
+      if not current_dir then
+         return nil
       end
+   end
+   return current_dir
+end
+local function is_in(a, b) -- a is in b
```
hishamhm, line 636: looks like this function is unused

hishamhm:
I'd like to get this merged as soon as possible since it's getting too big to do a review in one go. :)

I took another look at the PR (not a super detailed line-by-line look, I admit! it's a lot of stuff!) and left a couple of comments, apart from those I think we can get it merged now

tl
euclidianAce: I think I had a plan at one point to print warnings when generating files in the `source_dir` when `build_dir` was present. :shrug:

euclidianAce:
I'd say go ahead and merge this if you'd like. The most basic error cases have tests now and I've been using this since I made the pr on a bunch of tiny things and it hasn't had any trouble.

Here's a rundown for the changelog:
- `tl gen` now generates the file in the current directory, much like gcc and other tools
	- `-o --output` now lets you specify the output for a _single_ script
- `tl build` is a new command that will type check and compile a given set of files specified in `tlconfig.lua` of the current working directory or the command line. These new entries/flags are:
	- `-b --build-dir` `build_dir: string`: where generated files will end up
	- `-s --source-dir` `source_dir: string`: when provided, `build` will only include files from inside the given directory
	- `files: {string}`: a list of filenames to compile
	- `include` has been renamed to `--include-dir include_dir: {string}` to account for:
	- `include, exclude: {string}`: a list of patterns to find files to compile, what patterns are available are outlined in `docs/compiler_options.md`
- New flag `-p --pretend --dry-run`: don't write to any files, but type check and log what files would be written to.
- New config entry `skip_compat53: boolean` same as `--skip-compat53`

euclidianAce:
Hey @hishamhm, any updates on getting this merged?

hishamhm:
@euclidianAce hi, sorry for dropping the ball on this! I've been unable to put too much energy on Teal the last couple of weeks. I just merged it manually, squashing some of the commits together. Thank you for the nudge, this PR is a huge win for the project!

euclidianAce:
Thanks, no worries!

PR #184, opened 2020-07-06
rockspec: install both tl.lua and tl.tl

pdesaulniers:
By default, tl.tl was not installed, so we couldn't `require('tl')` inside external Teal code.

Does this tweak look right? My initial attempt was 
```lua
   modules = {
      tl = { "tl.lua", "tl.tl" },
   },
```
...but then, LuaRocks would try to compile these files with GCC.

### DIFF
--- a/tl-dev-1.rockspec
+++ b/tl-dev-1.rockspec
@@ -25,6 +25,9 @@ build = {
    install = {
       bin = {
          "tl"
+      },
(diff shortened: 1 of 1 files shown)


hishamhm:
I guess this works as a hack for the time being! LuaRocks could use some Teal support. ;)

PR #191, opened 2020-07-16
Record operator overload

### LINKED ISSUE #190: Record operator overloading
Are there any current plans for how to implement this? I've written up a very rough pr (no tests or anything) for this feature [here](https://github.com/euclidianAce/tl/tree/record_operator_overload) but would like some feedback/opinions on it first.
- First of all, syntax: this is what I settled on. The following is valid in the pr
```
local Vector = record
   {number}

   has +(Vector, Vector): Vector
   has /(Vector, number): Vector
   has ..(Vector, Vector): string

   has -(Vector, Vector): Vector

   -- technically unary ops arent in yet, but would be quite easy
   has #(Vector): number
   -- The framework is there to allow for binary and unary ops to have the same symbol
   has -(Vector): Vector

   -- etc

   -- has is still a valid name
   has: number
end

local a: Vector = {}
local b = a + a
local c = a / 4
```
- Later on some assertions can be added that at least one of the arguments contains the type but I want to stress that this is a proof of concept

- I'm haven't fully studied the parser yet but I was thinking a "has" fact and operator could be added and we could (as @hishamhm mentions in #97) use the generated table to store the information and define some reserved keys in tables. For example
```
local Vector = record has+(Vector, Vector):Vector end
local a: Vector = {}
if a has + then
   local b = a + a
end
```
would generate
```lua
local Vector = {["__teal_has_+"] = true}
local a = {}
if Vector["__teal_has_+"] then
   local b = a + a
end
```
(although I think lua reserves double underscore names)
and the compiler would error when someone tries to use `has` on a non record type

Speaking of #97, I think this is a decent way of adding operator overloading without stepping too far into committing to an OOP model.

notably this doesn't hook into metatables at all currently. My reasoning for this is that it's hard and I didn't want to get too far into anything without a second look at these ideas. Also metatable info is hard to track and can be (suprise suprise) super dynamic, not in practice usually, but it can be.

(Also I just realized I hadn't thought about generics at all ¯\\\_(ツ)\_\/¯)

euclidianAce:
Should close #190, I've written some basic tests for this, but generics are still a problem and I'm not 100% sold on the syntax. Additionally, unary operators shouldn't need their input's specified because they can't be called without the type being the argument. (binary ops need it because order and operations with other types will care)

So just a review of the issue I opened:
- syntax:
```
local Vector = record
   {number}

   has +(Vector, Vector): Vector
   has -(Vector, Vector): Vector
   has -(Vector): Vector
end

local a: Vector = {}
local b: Vector = a + a -- doesn't error! :D (at least at compile time)
```

still some opportunity for a 'has' operator, but I don't see much of a use case for it
```
if a has + then
   local b = a + a
end
```

But I do think this opens up some opportunity to do some refactors of how binary and unary operations are handled.

Additionally, the errors generated by overloading one operator multiple times are considered syntax errors which at first seemed wrong, but I guess makes sense.

Anywho, I'd like your thoughts on this. The syntax is easy to change and the logic was actually not too hard to implement (hopefully it actually works correctly).

(also generics are a little broken something like `has +(Vector<T>, T): Vector<T>` doesn't work due to the single `T` and I'm not exactly sure how those get resolved yet, furthermore stuff like `has +<T>(Vector, T): T should probably be allowed`)

### DIFF
--- a/tl.lua
+++ b/tl.lua
@@ -716,7 +716,7 @@ local table_types = {
    ["emptytable"] = true,
 }
 
-local HasType = {}
+local OpType = {}
 
 
 
@@ -1841,6 +1841,31 @@ local function parse_return(ps, i)
    return i, node
 end
 
+local unary_op = {
+   ["-"] = true,
+   ["#"] = true,
+   ["~"] = true,
+}
+local bin_op = {
+   ["+"] = true,
+   ["-"] = true,
+   ["*"] = true,
+   ["%"] = true,
+   ["/"] = true,
+   ["//"] = true,
+   ["^"] = true,
+   ["&"] = true,
+   ["|"] = true,
+   ["<<"] = true,
+   [">>"] = true,
+   ["=="] = true,
+   ["~="] = true,
+   ["<="] = true,
+   [">="] = true,
+   ["<"] = true,
+   [">"] = true,
+   [".."] = true,
+}
 parse_newtype = function(ps, i)
    local node = new_node(ps.tokens, i, "newtype")
    node.newtype = new_type(ps, i, "typetype")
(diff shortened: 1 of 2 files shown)


euclidianAce:
Just spitballing more ideas here:

One syntax that I did consider was just writing out the binary expression with the types.
For example:
```
local Vector = record
   {number}

   has Vector + Vector : Vector
   has Vector - Vector : Vector
   has Vector|number * Vector|number : Vector
   has Vector >> number : Vector
   -- etc.

   has # : number
end
```
or annotating the metamethod names instead of the operators themselves
```
local Vector = record
   {number}

   has __add(Vector, Vector): Vector
   -- etc.
end
```
or having a 'meta' block
```
local Vector = record
   {number}

   meta
      __add(Vector, Vector): Vector
      -- etc.
   end
end
```

hishamhm:
> or having a 'meta' block

This is the option most similar to what I've been thinking of, when thinking about adding metamethod support, though given the signatures of metamethods are standardized, not sure if it's necessary to type the arguments and returns. This is a longer conversation, because there are so many open questions:

* it needs to support both operators defined in Teal and records coming from Lua libraries: that is, the syntax should support both "interface-only" definitions as well as "interface-and-implementation". We don't allow record values declared inside `record` definitions (it is "interface-only"), so if we start using an "abstracted" syntax for the interface like, say, `has Foo + Foo: Foo` then it might be weird if the implementation is actually something way more "concrete" like `function Foo_mt.__add(a: Foo, b: Foo): Foo`
  * I think I prefer sticking to the concrete metamethod names
* how is the metatable declared in a Teal record with metamethods: to ensure type safety it would be better if it was done in an implicit way instead of having the programmer do the `setmetatable` dance themselves, but then whichever way the feature implements it becomes the "favored" way of doing metatables in Teal (is that even a problem?)
  * Once we start handling metatables, then we're pretty close to OOP. I've been thinking of taking on that sooner rather than later. PR #194 which I just submitted was written with that in mind, actually!

> still some opportunity for a 'has' operator

I probably don't want to go there, because it is very "structurally-typed" (type checking via checking the parts of types), while I'm trying to keep things more "nominally-typed" (for things that have names, type comparison should be essentially name comparison).

euclidianAce:
> though given the signatures of metamethods are standardized, not sure if it's necessary to type the arguments and returns

I would say in some cases it's necessary. For example, a `Vector` class would want most of its binary operations to be `function(Vector, Vector): Vector` but would want `__div` and maybe things like `__shr` and `__shl` to be component-wise operations with `function(Vector, number): Vector`. Maybe default for included binary operations of some record `Foo` to be `function(Foo, Foo): Foo` and let the programmer override the defaults if necessary?

> the syntax should support both "interface-only" definitions as well as "interface-and-implementation"

Ah, my idea was only for "interface-only". Truth be told the motivation for this implementation/syntax was so I could attempt to use lpeg with Teal so I just needed an interface for patterns.

In addition, I wanted to try to add this feature without committing to a specific OOP implementation for records to stave off that debate, but that's inevitable, so ¯\\\_(ツ)\_/¯.

PR #194, opened 2020-07-21
Type declaration changes

hishamhm:
Some thoughts on syntax changes in preparation for language additions:

* `local type Name = Type` for type declarations, no longer mixed with variable declarations. They always worked pretty differently from variable declarations (no `: Type` section, no multiple assignment, always auto-const).
  * `Type` there accepts any type, so you can give names to unions, etc. (so this not only a syntax change, but a new language feature)
  * remove `functiontype` as it becomes redundant, you just need to declare a `function` type as usual
* New shorthand forms `local record Name ... end` and `local enum Name ... end`, just like `local function Name...` exists in parallel to `local Name = function...` in Lua.
* both of the above for `global` as well, of course

### DIFF
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -28,13 +28,17 @@ precedence, see below.
        ‘for’ Name ‘=’ exp ‘,’ exp [‘,’ exp] ‘do’ block ‘end’ |
        ‘for’ namelist ‘in’ explist ‘do’ block ‘end’ |
        ‘function’ funcname funcbody |
-       ‘local’ ‘function’ Name funcbody |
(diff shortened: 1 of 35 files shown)


euclidianAce:
Huge :+1: to the `local record Foo` as I already do that accidentally.

euclidianAce:
Are there plans to allow generic types? For example, a set `{type:boolean}` is a really common pattern. Probably with syntax similar to: `local type Set = <T>{T:boolean}` or `local type Set<T> = {T:boolean}`

hishamhm:
> Are there plans to allow generic types

Not at this time! Haven't thought it through if it's a straightforward substitution or yet another Pandora's box :)

Issue #201, opened 2020-08-08
Pretty-printing tl source code for the purpose of refactoring/formatting


Ruin0x11:
I was wondering if it would be possible to have an option to change how `tl` outputs its AST, such that you could preserve the formatting and comments of the original file.

Here is my use case: I have a very large Lua project (~190K SLoC at last count) that I want to refactor in various ways, like renaming the fields of tables or moving the location of a module and updating all the calls to `require` project-wide to point to the new location. It's difficult to do these things without using macros or writing specialized programs to parse and modify the source code.

My idea would be using a separate program that leverages `tl` to parse the code into an AST with the extra information like types, applying a modification to AST nodes that satisfy a predicate, and writing the `tl` or Lua code back to a string. 

The issue is that refactoring code like this requires a specific property of the parser and formatter, namely that whitespace and comments are preserved. At the time I was wanting this none of the Lua parsers used by projects such as `luacheck` had this property, even though ones that did like `lib2to3` existed for other languages, so I had to write my own, named [`kotaro`](https://github.com/ruin0x11/kotaro), specialized for the purpose of source code refactoring. It sort of worked but the parser was buggy, it needed lots of tests and extra effort in order to be usable, and the performance wasn't that good from what I could accomplish by myself, and still some of the refactorings I wanted to do, like changing function signatures, would require a lot more effort to accomplish. But consolidating effort on `tl`'s parser would be more beneficial and it also comes with type information, which would open up a lot more possibilities for refactoring. There are only a few things missing from `tl` to accomplish what I'm thinking of:

1. Ability to pretty print the AST as the original `tl` source instead of just Lua.
2. Option to preserve whitespace and comments when parsing/converting.

Also, this would be nice to have:

3. A way of generating the dependency graph of a project directory (from what can be determined statically), to know what files need to be reparsed if a change is made.

Does this sound like a good idea?

hishamhm:
Hi there, thank you for all the feedback!

Here's my feedback on the items above

> Ability to pretty print the AST as the original tl source instead of just Lua.

This should be achievable through a new pretty-printer function, preferrably in a separate module once I split `tl.tl` in multiple modules with more stable APIs between them. I would prefer to keep the current code dump function more straightforward since it's used at require()-time.

> Option to preserve whitespace and comments when parsing/converting.

[...]

Issue #84, opened 2020-04-04
Implement a beautifier / formatter tool


pdesaulniers:
At some point, it would be nice to have a tool like [rustfmt](https://github.com/rust-lang/rustfmt) to format Teal code automatically.

This tool could probably be based on tl's parser.

cloudfreexiao:
I think this tool is nice ...
https://github.com/Koihik/LuaFormatter

ildar:
related: https://github.com/teal-language/tl/issues/201

alerque:
[StyLua](https://github.com/JohnnyMorganz/StyLua) is *far and away* the best Lua code formatter out there, and has *some* degree of configuration available. Obviously that doesn't extend to Teal yet, but the underlying parser [full-moon](https://github.com/Kampfkarren/full-moon) does already support Luau. It would not be a huge stretch to add Teal as an input option, and then extend StyLua for additional output options.

ALameLlama:
https://github.com/JohnnyMorganz/StyLua/issues/1039
https://github.com/Kampfkarren/full-moon/issues/351

Just for tracking

PR #203, opened 2020-08-09
Add Teal output option to tl.pretty_print_ast

Ruin0x11:
Related to #84 and #201.

Adds an `is_tl` option to `tl.pretty_print_ast()`, which will output the parsed AST as Teal instead of Lua.

The output of pretty printing the source code of the compiler itself as Teal typechecks successfully.

Also changes the output of `show_type()` to be closer to the input source code, like preserving the state of optional tuple parens.

Limitations:

- It does not preserve whitespace or comments.
- I might have missed a couple of language features.

### DIFF
--- a/tl.tl
+++ b/tl.tl
@@ -24,11 +24,11 @@ local tl = {
 -- Lexer
 --------------------------------------------------------------------------------
 
--- local inspect: function(value: any, options: InspectOptions): string = require "inspect"
-
 local inspect = function(x: any):string
    return tostring(x)
 end
+
+-- local inspect: function(value: any, options: InspectOptions): string = require "inspect"
 
 local keywords: {string:boolean} = {
    ["and"] = true,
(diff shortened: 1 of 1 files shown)


hishamhm:
@Ruin0x11 I have played with the topic of preserving comments today, and I've come to the conclusion that the re-dumping the AST is probably not the best approach for a formatter, because of the mixture of concerns: https://hisham.hm/2020/12/28/parsing-and-preserving-teal-comments/

Ruin0x11:
Ok, that makes sense.

I remember one time I was trying to make a Lua parser, one way to handle comments was by making a "concrete syntax tree" which was basically a wrapper around the lexer's output and attach the comment as a prefix to the next node that was lexed. For anything more complicated you'd convert the CST to an AST and discard all the comments. The CST was also where anything related to formatting would take place.

PR #206, opened 2020-08-09
Add `repl` option to compiler frontend

Ruin0x11:
Adds an option to start a Teal REPL from the command line.

Maybe we could have `tl` called with no arguments default to starting the REPL, like Python/stock Lua?

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,10 @@ Once `tl` is in your path, there are a few subcommands:
 * `tl gen module.tl` will check for syntax errors and
   generate a `module.lua` file in plain Lua with all type annotations
   stripped.
+* `tl repl` lets you try out Teal on the command-line through an interactive
+  REPL.
+* `tl build` will compile all Teal files in the current or specified directory,
+  according to the options in a `tlconfig.lua` file. See the documentation for more details.
 
 ## Documentation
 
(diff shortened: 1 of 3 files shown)


hishamhm:
Wow, impressive work so far!

Some bits of feedback:

I've noticed it doesn't pick up the Lua standard library from the environment on execution:

```
Teal REPL (0.7.1+dev)
> global x: number 
> x = 2
> print(x)
string["return pri"]:1: attempt to call a nil value (global 'print')
stack traceback:
	string["return pri"]:1: in main chunk
	[C]: in function 'xpcall'
	./tl:626: in method 'step'
	./tl:644: in method 'run'
	./tl:648: in main chunk
	[C]: in ?
> x
2
> local z: number
> z = 9
> z
9
> x
2
> x + z
11
> z + 9
18
> io 
string["return io"]:1: attempt to call a nil value (global 'tonumber')
stack traceback:
	string["return io"]:1: in main chunk
	[C]: in function 'xpcall'
	./tl:626: in method 'step'
	./tl:644: in method 'run'
	./tl:648: in main chunk
	[C]: in ?
```

Also, the `"return " .. ` trick (as seen in the `lua` REPL) probably needs a smarter construction to account for Teal's arity checks.

```
> io.write("Hello")
1 type error:
:1:1: excess return values, expected 1 got 2
>
```

Finally, `tl` has `compat53` available as a dependency, so no need to write code in Lua 5.1 style and pull in compatibility functions. Instead, feel free to write code in 5.3 style and require compat53 (preferrably with a pcall so that Lua 5.3 installations don't need to have it around).

This is an amazing feature! Really nice to see the `tl` CLI driver evolve!

hishamhm:
Revisiting this PR, the direction we ended up taking for the `tl` CLI is to keep it more as a backend tool, more focused as a compiler (`tl check`, `tl gen`) and feeder for tooling (`tl types`) rather than as a higher-level tool (hence deprecating `tl build` in favor of https://github.com/teal-language/cyan). As such, I think a REPL would make more sense today as a stand-alone tool. Would still love to see it happen, but I believe it should take shape in a different form than this PR's iteration. Thanks for understanding!

PR #208, opened 2020-08-10
config: fix source dir not doing anything without include patterns

### LINKED ISSUE #207: tl build: source dir is ignored when there are no include patterns

euclidianAce:
Fixes #207

### DIFF
--- a/tl
+++ b/tl
@@ -640,9 +640,9 @@ function project:find(path) -- allow for indexing with paths project:find("foo/b
    end
    return current_dir
 end
--- local function is_in(a, b) -- a is in b
---    return a:find("^" .. b) and true
--- end
+local function is_in(a, b) -- a is in b
+   return a:find("^" .. b) and true
+end
 
 project.dir, project.paths, project.emptyref = traverse(lfs.currentdir())
 project.source_file_map = {}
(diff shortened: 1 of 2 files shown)


tl
```diff
@@ -640,9 +640,9 @@ function project:find(path) -- allow for indexing with paths project:find("foo/b
    end
    return current_dir
 end
--- local function is_in(a, b) -- a is in b
---    return a:find("^" .. b) and true
--- end
+local function is_in(a, b) -- a is in b
+   return a:find("^" .. b) and true
```
hishamhm, line 644: If I'm following correctly, you're using this to mean that pathname `a` is within `b` (e.g. /x/y/foo.lua is in /x), right? This needs something after b to avoid having "/football is in /foo" == true

something like

```
local function is_in(a, b) -- a is in b
   local pattern = "^" .. b .. path_separator
   pattern = pattern:gsub(path_separator .. "+", path_separator)
   return a:find(pattern) and true
end
```

That gsub is not a complete normalization function, but should be good enough to work whether b contains a trailing slash or not. (LuaRocks has a bunch of those "path normalization" functions, I guess I should make that into a reusable library some time...)

euclidianAce: Ah, right. This was more of a hotfix as my idea with the `project` object was to basically be a wrapper for the filesystem. I think it would be a better idea to change the dir that `project:files` iterates over rather than doing some string comparison, which should be easy enough.

euclidianAce: This latest commit seems to fix this, and ive added a simple test case for it

hishamhm:
Merged!

PR #212, opened 2020-08-14
Allow enums to be indexed with string methods

### LINKED ISSUE #196: Allow enums to be indexed with string methods
This is probably just a subset of the bigger issue of type inheritance and such, but since enum members are strings, indexing them with string methods should be valid.
```
local Foo = enum
   "bar"
   "baz"
end
local var: Foo = "bar"
print(var:sub(1,1))
```
this results in:
```
cannot index something that is not a record: Foo
```

euclidianAce:
Allow enums to be indexed in the same way as strings since they are a subtype of strings + add a small test to check if this works for both enums and strings.

Closes #196


hishamhm:
@euclidianAce LGTM! Good call!

PR #215, opened 2020-08-26
Spec for `tl run` passing ... to target script

xvxx:
As per @hishamhm's comment in #214, here's a spec to test that `tl run` passes args through as `...` to the target script.


hishamhm:
@xvxx Thank you!!

Issue #216, opened 2020-09-05
Errors in modules uncaught by the tl compiler cause tl to error out


euclidianAce:
Minimal example:
```
-- my_module.tl
local function foo(...: any): function(): any
   return function(): any
      return select(1, ...) -- ... isn't allowed here
   end
end
return "hi"
```
```
-- my_script.tl
local mod = require("my_module")
print(mod)
```
which results in
```
$ tl run my_script.tl
lua: /home/corey/dev/tl//tl.lua:6525: bad argument #1 to 'concat' (table expected, got nil)
stack traceback:
        [C]: in function 'table.concat'
        /home/corey/dev/tl//tl.lua:6525: in function </home/corey/dev/tl//tl.lua:6507>
        [C]: in function 'require'
        my_script.tl:1: in main chunk
        (...tail calls...)
        [C]: in ?
```

what seems to be happening is that the module loading assumes that the tl compiler catches all errors and doesn't do any error handling for the `load` call, then falls through to a `table.concat` call on a nil table, since it found the module.
Should be a relatively easy fix. :)

PR #218, opened 2020-09-05
Add error checking for when tl doesn't catch lua errors

euclidianAce:
In addition, use `tl.load` to load modules since the code for both was very similar.

### DIFF
--- a/tl.tl
+++ b/tl.tl
@@ -12,7 +12,15 @@ local TypeCheckOptions = record
    result: Result
 end
 
+local LoadMode = enum
+   "b"
+   "t"
+   "bt"
+end
+local LoadFunction = functiontype(...:any): any...
+
 local tl = {
+   load: function(string, string, LoadMode, {any:any}): LoadFunction, string = nil,
    process: function(string, Env, Result, {string}): (Result, string) = nil,
    process_string: function(string, boolean, Env, Result, {string}, string): (Result, string) = nil,
    gen: function(string): string = nil,
@@ -6509,17 +6517,21 @@ local function tl_package_loader(module_name: string): any
    if found_filename then
       local input = fd:read("*a")
       fd:close()
-      local tokens = tl.lex(input)
-      local errs = {}
-      local i, program = tl.parse_program(tokens, errs, found_filename)
+      local errs: {ParseError} = {}
+      local _, program = tl.parse_program(tl.lex(input), errs, module_name)
+      if #errs > 0 then
+         error(module_name .. ":" .. errs[1].y .. ":" .. errs[1].x .. ": " .. errs[1].msg)
+      end
       local code = tl.pretty_print_ast(program, true)
-      local chunk = load(code, found_filename)
+      local chunk, err = load(code, module_name, "t")
       if chunk then
          return function(): any
             local ret = chunk()
             package.loaded[module_name] = ret
             return ret
          end
+      else
+         error("Internal Compiler Error: Teal generator produced invalid Lua. Please report a bug at https://github.com/teal-language/tl")
       end
    end
    return table.concat(tried, "\n\t")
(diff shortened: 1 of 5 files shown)


euclidianAce:
Fixes #216

tl.tl
```diff
@@ -6509,17 +6517,16 @@ local function tl_package_loader(module_name: string): any
    if found_filename then
       local input = fd:read("*a")
       fd:close()
-      local tokens = tl.lex(input)
-      local errs = {}
-      local i, program = tl.parse_program(tokens, errs, found_filename)
-      local code = tl.pretty_print_ast(program, true)
-      local chunk = load(code, found_filename)
+      local chunk, err = tl.load(input, module_name, "t")
```
hishamhm, line 6520: I'm not sure by just looking at the diff, but: shouldn't it be found_filename here instead of module_name? So that any parse errors contain the correct filename

euclidianAce: Ah, yes it should.

tl.tl
```diff
@@ -6509,17 +6517,16 @@ local function tl_package_loader(module_name: string): any
...
+      local chunk, err = tl.load(input, found_filename, "t")
       if chunk then
          return function(): any
             local ret = chunk()
             package.loaded[module_name] = ret
             return ret
          end
+      else
+         io.stderr:write("!!Error uncaught by tl!!\n", err, "\n\n\tThis is probably a bug caused by tl not catching invalid Lua.\n\tPlease consider reporting this at https://github.com/teal-language/tl\n\n")
+	 os.exit(1)
```
hishamhm, line 6529: On a second reading, I'm not sure this is right. When tl.load _does_ catch a parse error, it will return `nil, err` normally, so this not an "error uncaught by tl".

In the case of `type_check_and_load`, the situation is different: that "error uncaught by tl" message refers to code generated by tl from an AST produced by `tl.process` which finished without catching errors. (That is, an error-free AST shouldn't be generating code that the Lua VM fails to `load`).

Looking at the original code, perhaps what we really need is to handle the errors of `tl.parse_program` (Teal caught an error) and `load` (Lua caught an error that Teal has let slip by) separately.

euclidianAce: Ah, I see what you mean. Something like this should (hopefully) be right.
```lua
local function tl_package_loader(module_name: string): any
   local found_filename, fd, tried = tl.search_module(module_name, false)
   if found_filename then
      local input = fd:read("*a")
      fd:close()
      local errs: {ParseError} = {}
      local _, program = tl.parse_program(tl.lex(input), errs, module_name)
      if #errs > 0 then
         error(module_name .. ":" .. errs[1].y .. ":" .. errs[1].x .. ": " .. errs[1].msg)
      end
      local code = tl.pretty_print_ast(program, true)
      local chunk, err = load(code, module_name, "t")
      if chunk then
         return function(): any
            local ret = chunk()
            package.loaded[module_name] = ret
            return ret
         end
      else
         io.stderr:write("!!Error uncaught by tl!!\n", err, "\n\n\tThis is probably a bug caused by tl not catching invalid Lua.\n\tPlease consider reporting this at https://github.com/teal-language/tl\n\n")
         os.exit(1) -- maybe error here instead as I'm not sure how os.exit interacts with tests
      end
   end
   return table.concat(tried, "\n\t")
end
```
But I'm not too certain of a good way to test this without relying on Teal not behaving correctly.
Perhaps `tl.parse_program` could be overridden to never error and pass bad code to `load`?
Something like:
```lua
local tl = require("tl")
describe("Error checking", function()
   local old_parse_program
   setup(function()
      old_parse_program = tl.parse_program
      tl.parse_program = function(tokens, _, chunkname)
         return old_parse_program(tokens, {}, chunkname)
      end
   end)
   teardown(function()
      tl.parse_program = old_parse_program
   end)

   it("should report when Teal doesn't catch a Lua error", function()
      -- [[ ... ]]
   end)
end)

```

hishamhm: Agree that `error()` is better than `os.exit`. Being a "library" module, the `tl.tl` file must never `os.exit`.

About the testing, this sounds like a good strategy. But I wouldn't sweat a lot if this is too hard to test. It's very much a corner case. 

Also, I would change the error message to `Internal Compiler Error: Teal generator produced invalid Lua. Please report a bug at https://github.com/teal-language/tl`

euclidianAce:
Alright, error messages changed and a small test added. Seems to work on my end. Let me know if it needs anything else!

PR #224, opened 2020-09-13
Add testcases for __tostring metamethod

Zash:
Based on

https://github.com/teal-language/tl/blob/4d031432fe0d0823fdb72437b56d8581d4417af0/tl.tl#L3479

I'm assuming that a `__tostring` metatable field should not need to explicitly state that it returns a string, however this test case in this PR fails with

> excess return values, expected 0 got 1

### DIFF
--- a/spec/declaration/record_method_spec.lua
+++ b/spec/declaration/record_method_spec.lua
@@ -279,4 +279,21 @@ describe("record method", function()
       ]], output)
    end)
 
+   it("__tostring works with explicit return type", util.check [[
+         local foo_mt = record
+         end
+
+         function foo_mt:__tostring() : string
+            return ""
+         end
+      ]])
+
+   it("__tostring works without explicit return type", util.check [[
+         local foo_mt = record
+         end
+
+         function foo_mt:__tostring()
+            return ""
+         end
+      ]])
 end)


spec/declaration/record_method_spec.lua
```diff
@@ -279,4 +279,21 @@ describe("record method", function()
...
+         function foo_mt:__tostring() : string
+            return ""
+         end
+      ]])
+
+   it("__tostring works without explicit return type", util.check [[
+         local foo_mt = record
+         end
+
+         function foo_mt:__tostring()
```
hishamhm, line 295: This fails because it is not declared anywhere that this is actually a metatable, it just happens to be a record method that shares the name of a metatable, which is in no way reserved. 

The `__tostring` type definition you see in the `standard_library` is for a predefined `METATABLE` type<sup>*</sup> that name from the global namespace, but that is only used for things like the return of `getmetatable` or the argument type of `setmetatable`. Regular records don' t get any of that by default.

But even if you do

```
local foo_mt: METATABLE = {}
```

you won't be able to declare

```
foo_mt.__tostring = function()
   return "hello"
end
```

On the contrary, tl will complain about the returns _twice_: both in the `function()` line, because now it wants you to match the `METATABLE` declaration and explicitly say `: string`, and in the `return "hello"` line because it doesn't match the `function()` declaration (and when you add `: string` both errors go away). Teal doesn't have inference of return types (or arguments), the type signatures of all `function`s must be given.

(<sup>*</sup> I don't like having polluted the global namespace with that name, but it was a shortcut in early development and I still haven't come up with a better solution)

hishamhm: So, just to clarify, the first test passes but it has nothing to do with metatables, and the second test fails due to the non-inference of function returns, also not because of metatables.

For metatable-specific test cases, this would pass

```
local foo_mt: METATABLE = {}
foo_mt.__tostring = function(): string
   return "hello"
end
```

and this would fail (correctly)

```
local foo_mt: METATABLE = {}
foo_mt.__tostring = function(): string
   return "hello"
end
```

and I think it's okay (at least the compiler would tell you what to do correctly in that case by flagging that the return is missing).

Zash: Haha, not sure why I thought that a record would be considered a metatable.

Thanks for the clarification!

PR #225, opened 2020-09-16
fix file handle leak in tl.process

euclidianAce:
Found this when toying around with my own cli and testing a build with like 20k files of just `print` statements and `tl.process` apparently only closed the file handle when it failed to read.


hishamhm:
@euclidianAce Oops! Thank you! :grin:

PR #231, opened 2020-10-14
Add simple type checking for config entries

### LINKED ISSUE #230: tlconfig: include field is not type checked (maybe more)
if the include field is not an array of strings nothing is reported, this may happen with some other config entries related to bulid.

euclidianAce:
Fixes #230

### DIFF
--- a/tl
+++ b/tl
@@ -42,24 +42,39 @@ end
 -- FIXME
 local function validate_config(config)
    local valid_keys = {
-      build_dir = true,
-      exclude = true,
-      files = true,
-      include = true,
-      include_dir = true,
-      preload_modules = true,
-      quiet = true,
-      source_dir = true,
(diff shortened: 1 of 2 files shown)


tl
```diff
@@ -42,25 +42,40 @@ end
...
+      source_dir = "string",
+      skip_compat53 = "boolean",
    }
 
-   for k, _ in pairs(config) do
+   for k, v in pairs(config) do
       if not valid_keys[k] then
          print(string.format("Warning: unknown key '%s' in tlconfig.lua", k))
+      else
+         -- TODO: could we type-check the config file using tl?
```
hishamhm, line 60: Good question! Probably, at some point! But this PR is already a good improvement!!

hishamhm:
Merged, thanks!!

PR #243, opened 2020-10-18
fix /tmp/config_type_check not being removed

lenscas:
When running `make` I ran into the problem of the `type_check_spec` test to fail because the `/tmp/config_type_check` folder didn't get removed.

This was caused because `finally` only executes the last function that it was given.

Splitting the 2 tests up fixes this, as the cleanup function from the first test runs before the second overwrites it. It is now also possible to remove the `2` from the second folder, though I didn't do that in this PR (Can push that change as well if desired).

### DIFF
--- a/spec/config/type_check_spec.lua
+++ b/spec/config/type_check_spec.lua
@@ -1,7 +1,7 @@
 local util = require("spec.util")
 
 describe("config type checking", function()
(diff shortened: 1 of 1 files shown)


hishamhm:
Looks fine, thanks for the PR!!

PR #247, opened 2020-10-20
add start for build.tl support

lenscas:
Lots of things that need to happen still, but it is at least a start.

- [ ] move the generated code to the correct folder. 
- [ ] have a good way to tell the build script to stop compiling, as things are broken.
- [ ] keep track of when the build script ran 
- [ ] allow the build script to run code after the compile step. (I don't think this should happen with a `tl check`/ `tl run`. I'm not sure about gen)
- [ ] clean up old artefacts when build script needs to rerun.
- [ ] add automated tests
 
Right now, the main problem I'm running into is the fact that I can't find where teal stores the compiled version of the generated code. I guess it just keeps these in memory and I need to extract it manually somewhere?

Alternatively, instead of having build.tl generate teal code, it can generate lua code which can be moved directly to the correct location, but that doesn't feel right to me.

I'm now also using /tmp because teal doesn't have a set location to store build artefacts (Like how rust uses target to throw everything in). Does that need to change? (I at least think that how I get that path needs to change, because I doubt that that will work on windows....)

### DIFF
--- a/tl
+++ b/tl
@@ -51,6 +51,8 @@ local function validate_config(config)
       quiet = "boolean",
       source_dir = "string",
       skip_compat53 = "boolean",
+      build_file = "string",
+      build_file_output_dir = "string"
    }
 
    for k, v in pairs(config) do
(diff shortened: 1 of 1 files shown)


lenscas:
that the tests fail is already fixed on my end. Will push it later if I have more changes.

The problem was because I had `local build_path = ""` at line 284 and then proceeded to do simple checks like `if build_path then` which would always be true

lenscas:
closing this as it looks like github lost track of my fork (Hooray for me doing a bad thing when updating teal)

PR #251, opened 2020-10-25
lexer: fixes for the pretty-printer

Fang-:
"Various fixes and improvements." Most notably, it looks like words got renamed to identifiers, but the token pretty-printer (which... isn't presently used anywhere) never got updated to match.

Before:

```lua
localtypeatom=number

localtypecell=record
h:noun
t:noun
end

localtypenoun=atom|cell

functionhead(n:noun):noun
   ifniscellthen
      returnn.h
   else
      assert(false, 'head of atom; ' ..n)
   end
end
```

After:

```lua
local type atom = number

local type cell = record
   h: noun
   t: noun
end

local type noun = atom | cell

function head(n: noun): noun
   if n is cell then
      return n.h
   else
      assert(false, 'head of atom; ' .. n)
   end
end
```

### DIFF
--- a/tl.lua
+++ b/tl.lua
@@ -582,6 +582,8 @@ local add_space = {
    ["op:{"] = true,
    ["op:#"] = true,
 
+   ["::identifier"] = true,
+
    [",:identifier"] = true,
    [",:keyword"] = true,
    [",:string"] = true,
--- a/tl.lua
+++ b/tl.lua
@@ -626,6 +626,7 @@ local should_indent = {
    ["elseif"] = true,
    ["else"] = true,
    ["function"] = true,
+   ["record"] = true,
 }
 
 function tl.pretty_print_tokens(tokens)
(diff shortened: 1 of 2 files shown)


Fang-:
It's not obvious to me why just the luajit 2.1 CI is failing...

Fang-:
For the record, this isn't a complete solution as-is. The `keyword:(` causes function type definitions to be rendered as `local tail: function (n: noun)` even though you probably don't want that space there.

But also, the `function` in that context gets treated as a function definition for indentation purposes. Distinguishing between function and function-type definition in the lexer seems less straightforward a fix.

hishamhm:
@Fang- Looking good, thank you! Indeed, the lexer pretty printer was something made just as a development tool to debug the lexer, so it doesn't need to be perfect; as the language grows more complex I don't expect it to behave as well as the AST-based pretty-printer (though given the `x, y` information stored in the tokens, it could be possible to make a dump function that outputs something (nearly) identical to the input file — which wouldn't be a "pretty printer" anymore).

PR #254, opened 2020-10-30
tl.process: Cache required modules to not load code more than once

### LINKED ISSUE #245: tl doesn't remember 'in-project' files checked from `require`, resulting in code being loaded/checked more than once
the file structure is

```
tree
.
|-- tests
|   |-- test_cache.tl
|   `-- tests.tl
`-- tlconfig.lua
```

`tlconfig.lua`

```lua
return {
    include_dir = {
        "tests"
    }
}
```

`test_cache.tl`

```
global TestCache = {}

function TestCache.test_cache()
end
```

`tests.tl`

```
require("test_cache")
```

```
tl build
Wrote: tests/test_cache.lua
1 error:
tests/test_cache.tl:1:8: cannot redeclare global with a different type: previous type of TestCache is {test_cache: function()}
```

euclidianAce:
Fixes #245
Adds a `loaded: {string:Result}` field to `Env` to map filenames to results.

There ended up being some hitches trying to change `Env.modules` to map from filenames to `Result`s instead of the literal `require` argument to a type. So it ended up being simpler to add an `Env.loaded` field to map the filenames to the compiled results.

### DIFF
--- a/spec/api/process_spec.lua
+++ b/spec/api/process_spec.lua
@@ -20,4 +20,21 @@ describe("tl.process", function()
          util.assert_line_by_line(expected_lua_code, pretty_result_string)
       end)
    end)
+   describe("process", function()
+      it("should cache modules by filename to prevent code being loaded more than once (#245)", function()
+
+         local current_dir = lfs.currentdir()
+         local dir_name = util.write_tmp_dir(finally, "module_cache", {
+            ["foo.tl"] = [[ require("bar") ]],
+            ["bar.tl"] = [[ global x = 10 ]],
+         })
+
+         assert(lfs.chdir(dir_name))
+         local foo_result, foo_err = tl.process("foo.tl")
+         local bar_result, bar_err = tl.process("bar.tl", foo_result.env)
+         assert(lfs.chdir(current_dir))
+         assert(foo_result, foo_err)
+         assert(bar_result, bar_err)
+      end)
+   end)
 end)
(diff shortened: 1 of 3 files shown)


euclidianAce:
In the future I could see this breaking with some relative vs absolute filenames being the main issue, but the framework is there and that's a separate issue/pr imo.

hishamhm:
PR looks very clean, nice!

> In the future I could see this breaking with some relative vs absolute filenames being the main issue, but the framework is there and that's a separate issue/pr imo.

Good point. Normalizing can be tricky (esp. when things like Windows get involved) so let's punt that to some other time.

Since you were able to reproduce @virusdefender's in #245, do you think this one can fix the `tl build` behavior they were observing?

hishamhm (approved):

euclidianAce:
> Since you were able to reproduce @virusdefender's in #245, do you think this one can fix the tl build behavior they were observing?

The test that I added demonstrated the same behavior as #245 and this pr has that test passing, so at least on my end it seems to be fixed.

hishamhm:
@euclidianAce Ah, nice, I guess it was more of a "unit test" vs "integration test" question, as in: "the test checks that `tl.process` now does what we want, but does the fix propagate all the way so that the reproduction steps seen in #245 now start working or are there any other issues in `tl build`" — I wasn't really asking for an actual integration test, the unit test is fine, just curious if that `tl build` repro case from the issue is now known to be working.

euclidianAce:
Ah, yes (at least in my own local testing) reproducing the steps in #245 now doesn't error

hishamhm:
Awesome! I think we're looking good for a 0.8.2 release, I'll push that later today!

PR #255, opened 2020-11-04
tl: never return nil from stdlib function assert()

RyanSquared:
This makes it so the first argument passed to `assert()` can be ALPHA or nil, so the type of ALPHA can be considered not-nil by the fact that nil is a "non-alpha" case.

I have tested this locally (with code I'm actively writing) and it seems to work.

### DIFF
--- a/spec/stdlib/assert_spec.lua
+++ b/spec/stdlib/assert_spec.lua
@@ -1,0 +1,23 @@
+local tl = require("tl")
+local util = require("spec.util")
+
+describe("assert", function()
+   it("unwraps a specifically T|nil value into a T value", function()
+      util.mock_io(finally, {
+         ["assert.tl"] = [[
+            local function test(x: string): string
+               return x
+            end
+
+            local x: string|nil = "hello"
+
+            test(assert(x))
+         ]]
+      })
+      local result, err = tl.process("assert.tl")
+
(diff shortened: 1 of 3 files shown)


hishamhm:
@RyanSquared I'm still a bit confused by this :) Can you give a short example that demonstrates the difference in the new behavior? (even better if you can include it as a new file `spec/stdlib/assert_spec.lua` ;) )

RyanSquared:
will do

until that's done: I'd basically like it that I can self-describe a type T as `T | nil` (even though any T itself can be nil - this is for my own readability sake when looking at function signatures), then pass that to `assert(t)` where `t` is of type `T | nil`, returning type `T`.

Teal allows anything to be nil, as far as I can tell, but if you *specify* that a type can be nil, it has to be "unwrapped" in some manner, whether by checking `if t is T then`, or in shorthand by using `local x = assert(t)` which is an implied T (not specifically T|nil) from the current return value of `assert()`.

hishamhm:
As discussed on Gitter, we won't merge this, to avoid misunderstandings regarding the behavior of `T | nil` in Teal, since it is not the idiom to make types nilable as it would seem (all types are currently nilable).

This may be revisited if Teal gets a type system that is stricter wrt nils in the future!

Issue #198, opened 2020-07-27
Better error message for two types with the same unqualified name


Ruin0x11:
If you declare two types with the same name in different modules, but forget to use `as`, you get this error:

```
in return value: record field doesn't match: type: FileMode is not a FileMode
```

This is confusing. It would be better if the names were fully qualified instead, like this:

```
in return value: record field doesn't match: type: lfs.FileMode is not a love.filesystem.FileMode
```

hishamhm:
Fixed via #264!!

PR #258, opened 2020-11-08
tl: Initial exposure of types

euclidianAce:
This is just directly exposing the types as they are, but it definitely helps with cli drivers and things that want to work with the compiler programmatically in Teal itself.

Additionally at line 48 with
```
local Result = tl.Result
local Env = tl.Env
local Error = tl.Error
```

Causes a bunch of `Type is not Type` errors when done like
```
local type Result = tl.Result
local type Env = tl.Env
local type Error = tl.Error
```
Which could be the same cause as #198, but might be different?

Also the makefile failed at the `diff` step, since record fields don't get generated in order, hence the change at line 2643

### DIFF
--- a/tl.tl
+++ b/tl.tl
@@ -31,7 +31,7 @@ local record tl
    record Result
       ast: Node
       type: Type
-      syntax_errors: {ParseError}
+      syntax_errors: {Error}
       type_errors: {Error}
       unknowns: {Error}
       env: Env
@@ -698,13 +698,6 @@ local function new_typeid(): number
    return last_typeid
 end
 
-local record ParseError
-   y: number
-   x: number
-   msg: string
-   filename: string
-end
-
 local enum TypeName
    "typetype"
    "nestedtype"
@@ -971,7 +964,7 @@ end
 
 local record ParseState
    tokens: {Token}
-   errs: {ParseError}
+   errs: {Error}
    filename: string
 end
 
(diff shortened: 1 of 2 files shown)


tl.tl
```diff
@@ -2620,11 +2640,11 @@ function tl.pretty_print_ast(ast: Node, mode: boolean | PrettyPrintOpts): string
 
    local function print_record_def(typ: Type): string
       local out: {string} = { "{" }
-      for name, field in pairs(typ.fields) do
-         if field.typename == "typetype" and is_record_type(field.def) then
+      for _, name in ipairs(typ.field_order) do
```
hishamhm, line 2643: nice!

hishamhm:
> Which could be the same cause as #198, but might be different?

I'll take a look as soon as I can!

hishamhm:
I believe this PR has minor conflicts because #260 was merged, could you update it? Thanks!

hishamhm:
@euclidianAce I think the CI error stems from a bad forward declaration that had been fixed in #260, which your `record` declaration ended up resurrecting, but I haven't checked closely!

euclidianAce:
@hishamhm Ah, found it, it was an incorrect declaration of `tl.gen`, also I noticed that `ParseError` was the exact same type as `Error` and only used a few times, so I just did a quick search and replace on the type to only use `Error`

PR #259, opened 2020-11-08
Chore: make run_mock_project and tmp_dir no longer require a dir_name

euclidianAce:
Just to make test writing a bit easier for things like `build`.

They're not super useful except when temporarily replacing a `finally`
with a stub to see what went wrong, and a specific filename isn't
necessary for that and just makes writing the tests more annoying.


hishamhm:
Nice cleanup!

PR #260, opened 2020-11-11
Remove/fix shadow variable declarations

mingodad:
While trying to convert this project to LJS (https://github.com/mingodad/ljs) I was getting several errors related to redeclaration of variables (shadow variables) because in LJS it's an error to redeclare a variable in the same scope and a warning when shadowing a variable in outer scope.

And fixing this problem I found two interesting problems/bugs due to variable/type redeclaration:
```
local E: function(ParseState, number, Node): number, Node
...
local function E(ps, i, lhs, min_precedence)
```
```
local show_type: function(Type): string
...
local show_type: function(Type, {Type:boolean}): string
```
**I think that's a good idea to flag variable redecalration as an error !**

### DIFF
--- a/tl
+++ b/tl
@@ -107,7 +107,7 @@ local function get_config()
       config[k] = v
    end
 
-   local err = validate_config(config)
+   err = validate_config(config)
 
    if err then
       die("Error while loading config: " .. err)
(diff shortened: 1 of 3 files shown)


hishamhm:
Thank you for this PR! There are minor cosmetic things that I'd personally change but I think this can be merged as is!

> I think that's a good idea to flag variable redecalration as an error !

Right now I'd be more inclined to make it a warning!

Issue #257, opened 2020-11-06
Unable to use reserved keyword as a record field name


Ruin0x11:
It is not possible to use a reserved keyword like `end` as a record's field name, even using the syntax for such keywords from Lua.

```lua
local record Test
   ["end"]: number
end
```

```
4 syntax errors:
 test.tl:2:4: syntax error, expected identifier
 test.tl:2:5: expected a variable name
 test.tl:2:8: syntax error
 test.tl:3:1: syntax error
```

This is a problem when implementing network protocols that happen to use reserved Teal keywords in their JSON.

hishamhm:
Good catch — I guess the syntax you suggested above is the one we should adopt. PRs welcome if anyone wants to beat me to it! It should require only parser changes.

euclidianAce:
I have a local branch with the chage made and it works well, but what do you suggest for nested records/enums? Personally I think something like this:
```
local record MyType
   start: number
   ["end"]: number --works fine

   -- little awkward, but makes sense i guess
   type ["1stPart"] = record
      --[[ ... ]]
   end

   -- probably don't do this as the analog to tables would be allowing something like
   -- `function ["name"]() end` syntax in a table
   record ["1stPart"]
      --[[ ... ]]
   end
end
```

hishamhm:
> I have a local branch with the chage made and it works well

Awesome!

> what do you suggest for nested records/enums? 

Since type names are not Lua concepts mapped to Lua table keys, they don't need to support non-identifier names.

euclidianAce:
Okay then, I'll write up a couple tests for the branch and put up the pr later today.

hishamhm:
Fixed by @euclidianAce in #262!

PR #262, opened 2020-11-11
tl: allow arbitrary strings as record keys

euclidianAce:
- adds the same syntax for tables with arbitrary string keys to records, e.g.
```
record Foo
   ["end"]: number
   ['1stPart']: number
   [[[2ndPart]]]: number
end
```


euclidianAce:
Fixes #257

PR #271, opened 2020-12-02
add support for more gsub function argument return values

catwell:
The function argument of gsub can return a string, a number, `nil` or `false`.

This adds support for numbers and `nil`. I don't know how to encode `false` (I didn't want to add BOOLEAN which would incorrectly allow `true` but maybe I should).


hishamhm:
Thanks for contributing and welcome to Teal, @catwell! :smile: 

> I didn't want to add BOOLEAN which would incorrectly allow true but maybe I should

The type system does not distinguish between true and false, since it deals with types and not values, so yes, it's fine to use boolean when only false can be returned. :+1:

PR #274, opened 2020-12-05
add support for integer division

catwell:


hishamhm:
I wonder if the code generator should output `math.floor(a / b)` when running on Lua 5.1/5.2... Currently the bitwise operators (`|`, `&`, etc.) don't, so there's a precedent for just merging this PR as-is.

But taking the opportunity to brainstorm: do you think we should add this kind of smartness to the code generator, though? I already add `require("compat53")` to deal with standard library differences, so dealing with operator differences would be a logical next step (and being a source-to-source compiler, we can deal with those, unlike compat, which can only deal with library functions).

PR #275, opened 2020-12-06
Add the command to print the AST

Lurenlzm:
```
--test.tl

local function add(a: number, b: number): number
         return a + b
end

```
```

tl ast test.tl

```
Print the AST in JSON format .


hishamhm:
@Lurenlzm Thank you for the PR! I'm a _little_ reluctant about this because if people start depending too much on the format of the AST then it becomes harder to change it, but if people find this useful and accept that the AST may change over time, we can add this as an experimental feature.

euclidianAce:
I'm not a fan of adding this as it (in addition to what @hishamhm said)
 - is not hugely relevant for a cli tool, getting the ast of something is more of an api thing
 - adds another dependency
 - makes the cli a little more bloated, the api is not stable enough for the cli to survive as its own module, so keeping it as 1 lua script is (kind of) a requirement for now. Adding random features like this doesn't help that

and the amount of code added is small (~30 lines not counting tests) so I think it would make more sense just to include a small lua script in whatever you need this feature for. If you were planning on running teal to get the ast you already are able to run a lua script, so this shouldn't be an issue

hishamhm:
@euclidianAce Fair points! Let's close this one, and think of other possible APIs to provide for IDE tooling and the like.

euclidianAce:
I think the best thing for IDE support would be a language server. I have experimented with one and gotten it working, but for one file only and not a 'project'. I think a decent language server would need the cli to have an api for project management/traversal, which would make it more complex and probably worthy of being split off, which would require the `tl` api to be a little more stable and featureful like, for example, being able to query what variables are in a certain scope.

Do you think it would be worth it to open an issue/discussion about api design for stuff like this?

hishamhm:
I agree that a language server would be great, but we need to be careful to not fall into a trap of a rabbit hole of dependencies: "to have a language server, we need project support; to have project support, we need to handle dependencies; to have dependencies, we need a package manager; to have a package manager, we need an API; to have an API we need a stable API design; to have a stable API, we need to sort out the requirements..." ;) — got to be really careful otherwise it might take forever, never get finished or both :)

I'd much rather go the other way around and build it up from an MVP (which is the approach I've been taking for everything throughout this project :) ) and build up from there. @pdesaulniers was able to get the vscode integration started with the output of `tl check`. I think a good next building block would be:

> being able to query what variables are in a certain scope.

I've been thinking a lot about adding a command that outputs a list of file locations and types which could be consumed by editor plugins the same way they read `tl check`. Then it can get more sophisticated from there. And after things mature, sure, eventually we'll get the language server. :)

lenscas:
> we need to handle dependencies; to have dependencies, we need a package manager;

Can't we reuse luarocks for dependencies? Or are we not thinking of the same kind of dependencies/package manager?

Sure, a custom package manager for teal could do fun stuff (especially now that build scripts are a thing) but do we really want to build a one way gate between lua and teal?

euclidianAce:
> I'd much rather go the other way around and build it up from an MVP (which is the approach I've been taking for everything throughout this project :) )

Right now I'm working on a pr for lazy rebuilds but once I have that I can revise/rewrite the server that I had running as an mvp

Issue #273, opened 2020-12-04
Errors in patterns cause internal compiler error


catwell:
If you write:

    print((" xy"):match("[^\s]"))

instead of:

    print((" xy"):match("[^%s]"))

it causes:

> Internal Compiler Error: Teal generator produced invalid Lua. Please report a bug at https://github.com/teal-language/tl

hishamhm:
Thanks for reporting! This is an uncaught lexer error, this causes the same issue:

```
print("\s")
```

hishamhm:
The lexer does not validate escape sequences according to https://www.lua.org/manual/5.4/manual.html#3.1 — currently only detects a `\` when inside `singlequote_string` or `dblquote_string`, switches the lexer state machine to either `escape_singlequote_string` or `escape_dblquote_string`, which skip over the next character. Those states should add some checking that the character is a valid indicator, and produce an `$invalid$` token if the string is malformed (which will cause the parser to bail with `syntax error` later and avoid the error in `tl run`).

catwell:
I am taking a look at this, it is slightly more complicated than expected because of the escape sequences not composed of a backslash + a single character (backslash + line break, \z and unicode characters).

hishamhm:
Fixed by #276 :D

PR #276, opened 2020-12-06
add lexer code for string escapes that respects Lua 5.4 syntax

catwell:
This is an attempt at solving #273.

This PR is more to open a discussion than to be merged as is because it uses a helper function that looks ahead inside of the lexer.

I tried not doing like that (i.e. by adding extra states to the lexer) first but it adds a lot more code complexity.

(By the way, I add tl.lua in my PRs but please tell me if you prefer me not to.)

### DIFF
--- a/tl.lua
+++ b/tl.lua
@@ -197,6 +197,55 @@ local LexState = {}
 
 
 
+local escapable_characters = {
+   a = true,
+   b = true,
+   f = true,
+   n = true,
+   r = true,
+   t = true,
+   v = true,
+   z = true,
+   ["\\"] = true,
+   ["\'"] = true,
+   ["\""] = true,
+   ["\r"] = true,
+   ["\n"] = true,
+}
+
+local function lex_string_escape(input, i, c)
+   if escapable_characters[c] then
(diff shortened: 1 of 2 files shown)


hishamhm:
> This PR is more to open a discussion than to be merged as is because it uses a helper function that looks ahead inside of the lexer.

I think that's fine. If the lexer state machine used a stack of states, I guess this logic could be represented as regular states without a lot of duplication, but since it doesn't the helper is fine at keeping the end result simpler. Let's be pragmatic about it.

> (By the way, I add tl.lua in my PRs but please tell me if you prefer me not to.)

Yes, keep adding it, please! I know generated code is usually frowned upon in git repos, but I've decided to keep it for simplicity of bootstrapping by end-users. The current CI setup uses the tl.lua included in the PR, so it's best to include it.

Could you add a few tests in `spec/lexer` to exercise this code and trigger some syntax errors? Thanks!

hishamhm:
Merged manually! I also added some test cases and caught and fixed a small issue with numeric escapes (`print("hello\27hello")`

Thank you!

Issue #100, opened 2020-05-09
check argument number


virusdefender:
for example

```
local function a(i: number): number, number
    if i == 1 then
        return 1, 2
    else
        return 3
    end
end
print(a(1))
print(a(0))
```

```
Type checked main.tl
0 errors detected
```

hishamhm:
@virusdefender Thanks for the report! This is a tricky one.

This is very much related to #71, except that that one is about input arguments to a function, and this one is about output values.

The reason why Teal (currently) accepts missing return values is because any return value may be `nil` (because any type may be `nil`).

It is a very common pattern than functions in Lua will return sometimes one, sometimes two values:

```lua
function f(x)
   local result = x * 2

   if x < 10 then
      return nil, "oops"
   end

   return result
end
```

The "real" result type of `f` (the one that [lives in the Lua programmer's head](https://www.youtube.com/watch?v=OPyBQRndLUk)) is "either number, or a nil followed by a string". Note that this is different from `number, string` and from `number?, string?` (Teal doesn't even support `T?` yet). This is more like `(number) | (nil, string)`, which is a sort of "union type on implicit tuples". 

How to destructure this type on the caller gets pretty tricky. Common Lua pattern:

```lua
local x, y = f()
if not x then
   print("error: " .. y)
end
```

The above looks logical (if x is nil, then y is string), but it has a bunch of things to consider when making the compiler smart enough to understand this:
* what it if the valid result of f was not number but boolean? (`true` is a common success value)
* what if the valid result value was multiple values, as in (`(number, number) | (nil, string)`))
* what if other conditional assignments happen in between? (this needs to be a proper flow analysis) 

It goes on and on...

**Putting all that aside**, the simplest strict thing would be to force an explicit _arity_ on the return values. So, if the return is `number, number` and you want to return nothing on an error, you have to return `return nil, nil`.

This is the simplest to implement, but I think it would be pretty annoying to use, because considering error results of the type `nil, string`, this means that every function that returns errors becomes a multiple-return function, so you have to keep returning as `return result, nil` on every non-error return.

Another option, would be to do the same as what is considered on #71, and add notation not about nullability, but about _explicit arity_ (that is, the number of return expressions, regardless of their values and types).

One issue with that is syntactical: how to specify it? In #71 the `?` notation is added to the variable names. But there are no variable names in returns... The best I can think of is this:

```lua
function f(x: number): number, ?string
   local result = x * 2

   if x < 10 then
      return nil, "oops"
   end

[...]

Issue #71, opened 2020-03-15
Required / Optional function arguments?


pdesaulniers:
It would be nice if TL could prevent this kind of error somehow:

```lua
local function add(a: number, b: number): number
	return a + b
end

add() -- attempt to perform arithmetic on a nil value (local 'a')
```
Perhaps TL could implement TypeScript's syntax for optional parameters, and assume that non-optional parameters are required?
```lua
local function dumbadd(x: number, y?: number): number
    y = y or 0
    return x + y
end

dumbadd() -- error: missing argument 'x'
```

hishamhm:
I have thought about this too. The "correct way" to deal with this would be via nilable types (or rather, non-nilable types), but that's a bit of a [rabbit hole :)](https://www.youtube.com/watch?v=HfnjUCRzRKU). 

One possibility indeed is to have a much more restricted form, which is merely about the syntactic presence of arguments in function calls (in your example, as in TypeScript, the `?` is in `y?`, not in `number?`). I think it's worth giving this a try — I'll play a bit with this idea, thanks for the suggestion!

pdesaulniers:
> The "correct way" to deal with this would be via nilable types (or rather, non-nilable types)

Yeah, that would be great! `nil`-related errors are so annoying :) However, as you said, that does seem complex to implement.

Mehgugs:
The `?` syntax you mention is sugar for a union with `undefined`:
```ts
const example_1 = (x: number, y?: number) => x + (y || 0)
const example_2 = (x: number, y: number | undefined) => x + (y || 0) 
```

I think the current union type in teal can already work here?  All that remains is to add the "strict" mode similar to [typescript's nullable support](https://www.typescriptlang.org/docs/handbook/advanced-types.html#nullable-types)

Ruin0x11:
There is a lot more discussion about having nullable types in this TypeScript issue.

https://github.com/Microsoft/TypeScript/issues/13778

TypeScript does not make the return value of array accesses nullable, but this is a feature that many people still want. However it cannot be added to the language at this point without a complier flag, because it would break backwards compatibility.

In one of the FOSDEM talks for Teal the issue of having to make every array access nullable was mentioned.

I personally believe that making the return type of array accesses nullable early on will be better for type safety. The non-nullable types can be reserved for records. Some people say that not making array accesses nullable went against TypeScript's goal of preventing the user from forgetting things at compile time and not having them go looking for errors at runtime.

hishamhm:
[...]

PR #279, opened 2020-12-15
Allow some array literals in assignment to heterogenous tuples

euclidianAce:
Previously arrays would outright be incompatible due to the entire tuple
being compared, but since the type of array literals now carry their
length, we now only check up to that length for compatability (provided
that it's positive*)

e.g. Assignments like `local t: {number, string}
= { 1 }` are now valid, where they previously weren't

More formally: One can imagine an array as an infinite tuple of a single
type. To compare an array to a tuple is to truncate the array to a tuple
of matching size, then compare component-wise. If that array has a
known/inferred length, then both tuples can be truncated to that length.

* Arguably "0 length" arrays could be treated as length 1 as the sort of
  'minimum' requirement for an array to match a tuple is at least the
  first element must match

### DIFF
--- a/spec/declaration/tuple_spec.lua
+++ b/spec/declaration/tuple_spec.lua
@@ -17,16 +17,20 @@ describe("tuple declarations", function()
    }))
 
    it("should report when an array is of incorrect type of a tuple entry", util.check_type_error([[
-      local a: {number, string} = { [-1] = "hi" } -- infers to {string}
-      local b: {number, string} = { "hi" } -- infers to {string}
-      local c: {string, number} = { [-1] = "hi" } -- infers to {string}
-      local d: {string, number} = { "hi" } -- infers to {string}
+      local a: {number, string} = { [-1] = "hi" } -- infers to {string} with a length of 0, which is incompatible with {number, string}
+      local b: {number, string} = { "hi" } -- infers to {string} with a length of 1, which is incompatible with {number, string}
+      local c: {string, number} = { [-1] = "hi" } -- infers to {string} with a length of 0, which should maybe be compatible with {string, number}?
    ]], {
       { y = 1, msg = "in local declaration: a: tuple entry 1 of type number does not match type of array elements, which is string" },
       { y = 2, msg = "in local declaration: b: tuple entry 1 of type number does not match type of array elements, which is string" },
       { y = 3, msg = "in local declaration: c: tuple entry 2 of type number does not match type of array elements, which is string" },
-      { y = 4, msg = "in local declaration: d: tuple entry 2 of type number does not match type of array elements, which is string" },
    }))
+
+   it("should allow array literals that fit within then tuple length", util.check [[
+      local a: {number, string} = { 1 }
+      local b: {boolean, boolean, number} = { false, true }
+      local c: {string, number} = { "hi" }
+   ]])
 
    it("should report when a tuple has incompatible entries", util.check_type_error([[
       local b: {number, string} = { 1, false }
@@ -48,6 +52,7 @@ describe("tuple declarations", function()
    pending("should error with explicit integer indices that are out of range", util.check_type_error([[
       local c: {number, string} = { [-1] = 10 }
    ]], {
-      { msg = "in local declaration: c: got {number : number}, expected {number, string}" },
+      -- Something about the table being a map or about -1 being out of range
+      -- { msg = "in local declaration: c: got {number : number}, expected {number, string}" },
    }))
 end)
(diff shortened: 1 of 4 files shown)


hishamhm:
I think I would personally prefer the stricter behavior, but given the pervasive behavior of nils in Lua tables, I can see how someone would expect the behavior described in this PR.

But there's also often requests about making Teal stricter about detecting "missing data" in various ways — missing arguments in function calls (#71), missing expressions in `return` (#100). Restricting nil-ability of types is a big semantic change which could prove difficult, especially in interoperability with Lua libraries. But restricting the arity of composite data (arguments, returns, literals) would be much easier to do.

In that sense, in an ideal world I would prefer instead make tuples _stricter_ by requiring that literals assigned to tuples have the correct arity, e.g. `local T:  {number, string, string} = {10, nil, nil}`. Since tuples are a new concept in the language, it could be a good idea to avoid the footgun of forgetting a tuple element from the get-go.

(If writing down the explicit nils are so problematic, one could also consider an optional-arity marker for things like `local T:  {number, ?string, ?string} = {10}` similar to what was discussed in #71 and #100, but I'm still concerned about the potential confusion between arity and nilability that might arise, especially in things like return types and tuples — in function arguments maybe it wouldn't be so bad because they would be in the variable and not the type.)

As the current code in `master` stands, we're right now in a halfway between being lax as this PR proposes and strict as I described in the comment above, because `local t: {string, number} = { "hi" }` fails but `local t: {string, string} = { "hi" }` doesn't — so we don't really have proper arity checks in place.

euclidianAce:
I'm personally not a huge fan of having to explicitly write trailing `nil`s, but I do see the benefit of being explicit about it. For 0.9, you can keep the current behavior if you like and maybe this can be a discussion ground for some stricter arity/nilablility checks. (or maybe enable githubs discussion feature thing for this repo if you can? idk how github rolls out beta stuff but I've been able to do it on a couple of my repos)

As far as arity goes, is there any benefit for making some special syntax for annotating it? like `local t: {string, number}#1 = { "hi" }` says the tuple has minimum length of 1? I'm not really a fan of this as the type annotation is starting to look cluttered but it could have its uses? like maybe having a length 'hint' for arrays, (which i guess would just be a shorthand for a homogenous tuple)

I don't really have a super strong preference on the behavior, but I'd lean more towards being explicit.

spec/declaration/tuple_spec.lua
```diff
@@ -48,6 +52,7 @@ describe("tuple declarations", function()
    pending("should error with explicit integer indices that are out of range", util.check_type_error([[
       local c: {number, string} = { [-1] = 10 }
    ]], {
-      { msg = "in local declaration: c: got {number : number}, expected {number, string}" },
+      -- Something about the table being a map or about -1 being out of range
+      -- { msg = "in local declaration: c: got {number : number}, expected {number, string}" },
```
hishamhm, line 56: What does the code above produce, with this PR?

hishamhm:
The current behavior in `master` is the weirdest as it is neither properly lax or properly strict with regard to arities. Since Teal is currently lax in other places, I'd rather merge this in now, and add the arity or nullability checks in tuples and other places later, if we choose to go down that route.

spec/declaration/tuple_spec.lua
euclidianAce: It currently produces the error
`'in local declaration: c: tuple entry 2 of type string does not match type of array elements, which is number'` which seems fine given that that's inferred as `{number}`. I think I marked that as pending because of how i wanted to change tables of that form to maps rather than arrays in the original pr.

euclidianAce: I definitely would prefer something saying something like `-1 is out of range for the tuple {number, string}`, but I think that would require types inferred from literals to carry more information than they currently do iirc

hishamhm: > i wanted to change tables of that form to maps rather than arrays in the original pr.

Ah yes, I changed that bit of "infer as map" but ended up not discussing it much. I'm generally trying to avoid surprises to Lua programmers by not assigning two different semantic meanings to two syntactic constructs that happen to be syntax-sugar to each other in Lua. (But I did break that guideline on occasion: function declarations with `:` keep the "I'm a method" metadata and are treated differently).

hishamhm: I'll remove this test case when merging the PR, because it's equivalent to the `c` case in line 26 :+1:

hishamhm:
Merged it manually with minor tweaks!

Issue #89, opened 2020-04-19
Should start to output warnings


hishamhm:
For things which are valid Lua and not type errors, but still are most likely bugs, such as shadowing variables (especially within the same scope) — the kind of warning we usually get from luacheck.

euclidianAce:
is #284 enough to close this issue? I think new issues for specific warnings/checks can be opened if needed

hishamhm:
Yes, fully agree!

PR #284, opened 2020-12-22
Implement warnings

euclidianAce:
A basic start on a warning system. Currently the only warnings implemented are for unused locals whose names aren't prefixed with an underscore (Stolen from Erlang, but I know there are other languages that ignore _ in similar manners, but not exactly in the same way), and redefined locals. So
```
local x = 10
print("hello world")
```
produces the result
```
$ tl check foo.tl
1 warning:
foo.tl:1:7: unused variable x: number
Type checked foo.tl
0 errors detected -- you can use:

   tl run foo.tl

       to run foo.tl as a program

   tl gen foo.tl

       to generate foo.lua

```
and editing it to
```
local _x = 10
print("hello world")
```
produces no warnings.

Additionally, unused arguments and unused functions will be annotated as such
```
local function foo()
end
```
results in
```
$ tl check foo.tl
1 warning:
foo.tl:1:1: unused function foo: function()
```
and
```
local function foo(x: number)
end
foo(10)
```
results in
```
$ tl check foo.tl
1 warning:
foo.tl:1:20: unused argument x: number
```

and
```
local a = 10
local a = 12
```
results in
```
$ tl check foo.tl
2 warnings:
foo.tl:2:7: redeclaration of variable 'a' (originally declared at 1:7)
foo.tl:2:7: unused variable a: number
```

Additionally, these types of warnings are very nice for a language server/linter

### Internal changes
 -  `node_warning` function that acts similarly to `node_error`
 -  add `warnings: {Error}` field to `Result`
 - `find_var` now returns the `Variable` that was found and `find_var_type` uses `find_var` and returns the type and `is_const` to replace the old behavior of `find_var`
 - adds fields to `Variable`
    - `declared_at: Node`
    - `used: boolean` - whenever a `Variable` is fetched using `find_var` this is set to `true`, when a scope is being closed, it checks this field for all the variables in it and generates warnings for unused ones 
    - `is_func_arg: boolean` - this one I'm not so sure is necessary, but was the easiest way to implement it. Maybe it could be useful for other things?
 - I went through `tl.tl` and eliminated all the warnings generated, hence the large diff :laughing:

### DIFF
--- a/tl.tl
+++ b/tl.tl
@@ -5286,6 +5286,7 @@ tl.type_check = function(ast: Node, opts: TypeCheckOptions): {Error}, {Error}, T
       for name, var in pairs(vars) do
          if var.declared_at
             and not var.used
+            and not var.is_narrowed
             and name:sub(1, 1) ~= "_"
          then
             node_warning(
--- a/tl.tl
+++ b/tl.tl
@@ -2218,7 +2218,7 @@ local function parse_variable_declarations(ps: ParseState, i: number, node_name:
    if #asgn.vars == 0 then
       return fail(ps, i, "expected a local variable definition")
    end
-   local lhs: Node = asgn.vars[1]
+   -- local lhs: Node = asgn.vars[1]
 
    i, asgn.decltype = parse_type_list(ps, i, "decltype")
 
@@ -2424,7 +2424,7 @@ local function visit_before<K,N,T>(ast: N, kind: K, visit: Visitor<K,N,T>): T
    end
 end
 
-local function visit_after<K,N,T>(ast: N, kind: K, visit: Visitor<K, N,T>, xs: {T}): T
+local function visit_after<K,N,T>(ast: N, kind: K, visit: Visitor<K,N,T>, xs: {T}): T
    if visit.after and visit.after.before then
       visit.after.before(ast, xs)
    end
@@ -2453,7 +2453,7 @@ local function recurse_type<T>(ast: Type, visit: Visitor<TypeName, Type, T>): T
    end
 
    if ast.types then
-      for i, child in ipairs(ast.types) do
+      for _, child in ipairs(ast.types) do
          table.insert(xs, recurse_type(child, visit))
       end
    end
@@ -2552,7 +2552,7 @@ local function recurse_node<T>(ast: Node,
          cbs.before_statements(ast, xs)
       end
       xs[2] = recurse_node(ast.thenpart, visit_node, visit_type)
-      for i, e in ipairs(ast.elseifs) do
+      for _, e in ipairs(ast.elseifs) do
          table.insert(xs, recurse_node(e, visit_node, visit_type))
       end
       if ast.elsepart then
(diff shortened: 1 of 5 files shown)


agladysh:
Isn't detecting unused variables something better left to a static linter like luacheck?

euclidianAce:
There is at least some desire for it (#89) and it's quite easy to implement given that all the code analysis/lexing/parsing is already implemented so i dont see much of a point of reimplementing it in a separate tool

lenscas:
there are also more and more languages that have it build in (Rust is another language where the compiler does that check).

tl.tl
```diff
@@ -4215,10 +4217,11 @@ local function add_compat53_entries(program: Node, used_set: {string: boolean})
...
 
-   for i, name in ipairs(used_list) do
-      local mod, fn = name:match("([^.]*)%.(.*)")
-      local errs = {}
-      local text: string
+   for _, name in ipairs(used_list) do
+      -- None of these are used?
+      -- local mod, fn = name:match("([^.]*)%.(.*)")
+      -- local errs = {}
+      -- local text: string
```
hishamhm, line 4224: good catch! those are left overs, they can go away!

tl.tl
```diff
@@ -4895,9 +4898,9 @@ tl.type_check = function(ast: Node, opts: TypeCheckOptions): {Error}, {Error}, T
       return known_table_types[t.typename]
    end
 
-   local function resolve_union(typ: Type)
-      assert(typ.typename == "union")
-   end
+   -- local function resolve_union(typ: Type)
+      -- assert(typ.typename == "union")
+   -- end
```
hishamhm, line 4903: This can be deleted as well!

tl.tl
```diff
@@ -2217,7 +2218,7 @@ local function parse_variable_declarations(ps: ParseState, i: number, node_name:
    if #asgn.vars == 0 then
       return fail(ps, i, "expected a local variable definition")
    end
-   local lhs: Node = asgn.vars[1]
+   -- local lhs: Node = asgn.vars[1]
```
hishamhm, line 2221: I think this is a left over from when type declarations were done with plain `local` and can go away too

hishamhm:
@euclidianAce this is awesome! The implementation looks clean, I read the commits in order and they were very easy to follow through. My only comments on it were all "yes, you can delete this bit of commented out code". :)

@agladysh I'll just echo what others said: I think this is fine because reporting unused variables is pretty standard for compiler warnings nowadays. I think it's more like luacheck over-extended into doing "typical compiler stuff" as well because, well, Lua doesn't have a compilation step. :) It's really cool how many warnings luacheck could do, short of performing actual type checking — it even includes some flow analysis which to this day we still don't have yet. But sadly, luacheck is currently unmaintained, so I wouldn't mind to eventually inherit more of its warnings into the compiler (but leave out things like [formatting issues](https://luacheck.readthedocs.io/en/stable/warnings.html#formatting-issues-6xx), which are "typical linter stuff").

hishamhm:
@euclidianAce I went ahead and deleted these commented out bits, added a commit for reporting unused goto labels, and merged! \o/

catwell:
re. luacheck: it is also somewhat annoying to use with tl because tl produces code that doesn't pass luacheck by default (in particular it defines Lua variable for types...).

PR #286, opened 2020-12-23
Fix warnings for 'unused label' on unresolved labels

euclidianAce:


euclidianAce:
also don't report internal @ variables for being unused

hishamhm:
oops, missed the other case! thanks for fixing this!

PR #295, opened 2020-12-26
feat: gh actions playground

### LINKED ISSUE #291: teal-playground bot previews
I don't know how much the `teal-language/teal-playground` is used to test tl during development, if at all, but would be cool to have an auto build env for PR's so that users can play around with the feature before it lands on master. The implementation would be dead simple since teal-playground is completely static and relies on a publicly accessible tl.lua. So every PR could `/s/master/<branch>`

Replace this line with the PR permalink:
https://github.com/teal-language/teal-playground/blob/040e5962c1eb839087b6445f8af37a8dba3a1fe9/src/components/Playground.vue#L39-L48

I think this has implications that we'd need to build a teal bot so opening up an issue to gauge interest and solicit feedback.

Typescript has a bot that builds playgrounds (among other things), example here: https://github.com/microsoft/TypeScript/pull/42020#issuecomment-748259864

darrenjennings:
Adds testing playground links for teal-language/tl pull requests using github actions + netlify. The tl;dr of the workflow file is:

1. git clone `teal-language/teal-playground`
2. build playground website with link to `.tl` source files in referenced PR as envs
3. deploy the site to a publicly accessible URL
4. comment on the PR with the URL to the deployed site

Fixes #291

### DIFF
--- a/.github/workflows/playground.yml
+++ b/.github/workflows/playground.yml
@@ -1,0 +1,57 @@
+name: playground
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [12.x]
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        repository: 'teal-language/teal-playground'
+        ref: 'master'
+        path: 'tl/teal-playground'
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: install
+      run: |
+        cd ${{ github.workspace }}/teal-playground
+        yarn install
+    - name: build
+      run: |
+        echo "${{ github.event.pull_request.head.repo.full_name }}/${{ github.head_ref }}"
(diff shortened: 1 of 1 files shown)


hishamhm:
@darrenjennings Nice! I think it's looking good to merge this and see what happens when we get PRs coming from other forks.

lenscas:
I don't think it will work, as secrets are not shared to actions running due to a fork: https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow

darrenjennings:
@lenscas I think we can use "pull_request_target" as the event to resolve this https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

> In order to solve this, we’ve added a new pull_request_target event, which behaves in an almost identical way to the pull_request event with the same set of filters and payload. However, instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the pull request. This means the workflow is running from a trusted source and is given access to a read/write token as well as secrets enabling the maintainer to safely comment on or label a pull request.

darrenjennings:
@hishamhm the change to the workflow file to `pull_request_target` means that it won't run on any PR's until this `.github/workflows/playground.yml` is on `master`.

PR #296, opened 2020-12-27
spec/util.lua: add batch_assertions()

euclidianAce:
A simple object to defer assertions so that other assertions may also be
run and so you can see each assertion that failed instead of having to
address one assertion, rerun busted, address another single assertion,
ad nauseam.

First a batch is created by calling `batch_assertions`, then with the
`:add(...)` method, you provide an assertion function and the parameters to
call with it. Then, at some point, call the `:assert()` method to run
each assertion.

Each assertion is run even if ones before it failed. The error objects
are `tostring`ed (since luassert does some fancy stuff with error
objects) and concatenated together in a nicely indented fashion to
distinguish between different assertions.

### DIFF
--- a/spec/util.lua
+++ b/spec/util.lua
@@ -3,16 +3,18 @@ local util = {}
 local tl = require("tl")
 local assert = require("luassert")
 local lfs = require("lfs")
-local current_dir = lfs.currentdir()
+local current_dir = assert(lfs.currentdir(), "unable to get current dir")
 local tl_executable = current_dir .. "/tl"
 local tl_lib = current_dir .. "/tl.lua"
+
+local t_unpack = unpack or table.unpack
 
 function util.do_in(dir, func, ...)
    local cdir = assert(lfs.currentdir())
    assert(lfs.chdir(dir))
    local res = {pcall(func, ...)}
    assert(lfs.chdir(cdir))
-   return (unpack or table.unpack)(res)
+   return select(2, assert(t_unpack(res)))
 end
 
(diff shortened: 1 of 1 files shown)


euclidianAce:
Short example since it isn't really reflected by code that passes tests:
```lua
local util = require("spec.util")

it("should fail", util.check_warnings([[
local x = 10
local x = 12
]], {
   { y = 1, msg = "redeclared variable x" }
}))

```
```
Failure → spec/blah_spec.lua @ 3
should fail
./spec/util.lua:73: batch assertion failed:


   Expected same amount of warnings:
      Expected objects to be the same.
      Passed in:
      (number) 3
      Expected:
      (number) 1

   [1] Expected same y location:
      Expected objects to be the same.
      Passed in:
      (number) 2
      Expected:
      (number) 1

   [1] Expected messages to match:
      Expected strings to match.
      Passed in:
      (string) 'redeclaration of variable 'x' (originally declared at 1:7)'
      Expected:
      (string) 'redeclared variable x'

```
whereas on the current master:
```
Failure → spec/blah_spec.lua @ 3
should fail
./spec/util.lua:369: Expected same amount of warnings:
Expected objects to be the same.
Passed in:
(number) 3
Expected:
(number) 1
```

hishamhm:
Ah, this is a much fancier version of the hack I had added into another PR!
https://github.com/teal-language/tl/pull/292/commits/7f1c361ca3356bf03b87d842f150a2c226d5028f

Issue #285, opened 2020-12-23
some bugs about the new `warning` feature


virusdefender:
@euclidianAce 

## case 1

```
local os = require("os")
print(os.clock())
```

> redeclaration of variable 'os'

there is no another `os` variable?

## case 2

```
local f = function(event: number)
    if event == 1 then
        goto break_restart
    else
        return
    end
    
    ::break_restart::
    print("break")
end

print(f(1))
print(f(2))
```

> test.tl:3:9: unused variable @unresolved: table: 0x55b6e9883a90
> est.tl:3:9: unused variable @unresolved: table: 0x55b6e9883a90
> test.tl:8:5: unused label ::break_restart::
> test.tl:3:9: unused variable @unresolved: table: 0x55b6e9883a90

euclidianAce:
thanks for the report, will look into these when I get the chance!
> there is no another os variable?

Technically there is the global `os` variable that you shadow with the local one, but this is a decently common pattern that this should probably be an exception. A small workaround for now if you want to get rid of the warning would just be to make a local of the function you need like so:
```
local os_clock = os.clock
print(os_clock)
```

As for the second case, I'm not exactly sure how labels are resolved, but its hopefully an easy solution. (pinging @hishamhm since you added the warnings for labels)

euclidianAce:
okay, #286 should solve that second case. The first one is a bit more complex of an issue that might warrant some discussion

virusdefender:
another case 

```
local start,  _, _ = string.find(s, sub, 1, true)
```

```
redeclaration of variable '_' (originally declared at xx:xx)
```

should it be accepted?

it's allowed in golang

```
package main

import "fmt"

func a() (int, int, string) {
	return 1, 2, "test"
}

func main() {
	val, _, _ := a()
	fmt.Println(val)
}
```

euclidianAce:
I don't know much about golang, but after a cursory glance the compiler throws an error about `a` returning 3 values so you need placeholder values there to acknowledge the arity of the function. In Lua you can just ignore those values without a placeholder
```lua
local function a()
   return 1, 2, "test"
end

local val = a()
print(val)
```
In fact it appears that golang is one of many languages that actually treats `_` as sort of a black hole that is used to acknowledge the existence of values, but not use them. Whereas in languages like Lua (and consequently Teal) just use `_` as convention for something to be ignored, e.g. in Go:
```go
package main
import "fmt"
func main() {
	_ := 10
	fmt.Println(_)
}
```
[...]

PR #300, opened 2020-12-28
do not emit a warning when localizing a global

hishamhm:
Localizing globals is a common Lua pattern (for better or worse...)
so let's not emit a warning when we detect that it is happening.

For simplicity, I don't check that the variable being aliased with
the same name is really a global, but that shouldn't matter.

This does not check that the dynamic _value_ is the same, only that the static
_name_ is the same. Therefore, people shouldn't do this to localize the
standard library: `local os = require("os")` but instead do this `local os =
os` (though it's often not needed because the Teal compat53 code generator
already does that for the standard library modules in use).

See #285.


euclidianAce (approved):
Looks good to me!

PR #301, opened 2020-12-29
start of the build.tl file

lenscas:
Lots of things that need to happen still, but it is at least a start. I think I am going to split it up as multiple pr's to reduce the amount of code changes in 1 pr.

First PR: The absolute basics.

- [x] move the generated code to the correct folder. 
- [x] get a temporary folder in a way that windows likes as well.
- [x] clean up after it is done ?

Second PR small improvements:
- [ ] give more information to gen_code than just the folder for generated .tl files. (Location for generated .lua files and maybe the project folder itself?)
- [ ] allow the build script to run code after the compile step. (Probably only for `tl build`)
- [ ] have a good way to tell the build script to stop compiling, as things are broken. (Just return a boolean + error message?.)

Third PR cache the results: 

- [ ] keep track of when the build script ran 
- [ ] clean up old artifacts when build script needs to rerun.


lenscas:
should a global type be exposed to help you write the `build.t`l file, so you can't make a typo in one of its keys?

One one hand I am in favor of it as more types is more better, on the other hand it is a "magic" type coming out of nowhere specifically for build scripts. Which I'm not a favor of.

hishamhm:
@lenscas What are the keys? Do you have an example of a build.tl file? (This PR draft is not including your tests yet.)

lenscas:
right now it is only "gen_code" (which I'm not super happy about) which is a function taking a string. The string is the path that it should put any teal files in it generates. Maybe a second parameter as well for files that should still be in the output but are not teal files?

Later it will also include a function that runs after teal is done. Which takes the build_dir as a parameter.

And assuming the "gen_code" function will get cached (Which I think is worth while in this case) it also will need to include an array of string patterns.

So, right now an example is
```lua
return {
    "gen_code" = function(path:string)
        local file = io.open(path .. "/file_to_generate.tl","w")
        file:write([[return "Hello from file generated by build.tl"]])
        file:close()
    end
}
```

Also, I think I found a cleaner solution instead of a type that gets injected out of nowhere. Just have a separate library that contains the type + a few helper functions that get automatically installed with teal. (A helper function could be a way to concatenate the path).

This library could then also include variants of functions from lua's std that are easier to work with as they take teals types into account (So, less if this string got passed, return type A, else return type B,etc. Instead have it more split up) and any base type/function that seems useful enough but is not included in lua's std.

lenscas:
I'm going to wait with the tests until #304 gets merged, as it is somewhat of a pain for me to test otherwise.

hishamhm:
#304 is merged :grin:

euclidianAce:
Just a couple things before I give this a more careful review
 - Few prs just got merged, would you mind rebasing it? I think the only actual change that affects anything is a new `report_result` function in `tl`
 - Could you write up some documentation covering the basic use case for a `build.tl` file and how to use it? I only kind of understand the intent behind it for generating code before a build for stuff like autogenerated typedefs and stuff like that. So just a simple explanation and use case would be good as I don't think merging a new feature without documentation is a good idea.

lenscas:
For some reason my laptop has 1 failing test even without my code, so I am not sure what is up with that.

As a result I decided to ignore it and push anyway as it may just be my laptop?

Failing test:
```
Failure → spec/cli/check_spec.lua @ 19
tl check on .tl files reports number of errors in stderr and code 1 on type errors
spec/cli/check_spec.lua:31: Expected strings to match.
Passed in:
(string) 'lua: attempt to index a nil value
stack traceback:
        [C]: in for iterator 'for iterator'
        ./tl:342: in upvalue 'report_warnings'
        ./tl:432: in upvalue 'report_result'
        ./tl:454: in local 'type_check_file'
        ./tl:550: in main chunk
        [C]: in ?
'
Expected:
(string) '2 errors:'

stack traceback:
        spec/cli/check_spec.lua:31: in function <spec/cli/check_spec.lua:19>
 ```

**Comment by hishamhm** [comment]
I ended up breaking `master` by merging multiple green PRs without testing CI in between merges. I'm pushing the fix!

**Comment by hishamhm** [comment]
@lenscas `master` should be green now, if try to checkout again

**Comment by lenscas** [comment]
it already passed CI though? There are also no merge conflicts.

File: tl [hunk 1]
```diff
@@ -498,6 +621,23 @@ local function type_check_and_load(filename, modules)
    return chunk
 end
 
+--------------------------------------------------------------------
+--                        RUN BUILD SCRIPT                        --
+--------------------------------------------------------------------
+local script = {}
+if tlconfig.build_file then
+   script = type_check_and_load(tlconfig.build_file,modules)()
```
euclidianAce, line 629: This should probably be
 - `pcall`ed to gracefully error out if there's an error in `build.tl`
 - manually type checked to, again, gracefully error out on user error

Basically we don't want user error to make a stacktrace point to inside of `tl`

tl
```diff
@@ -787,6 +873,14 @@ function project:files(inc_patt_arr, exc_patt_arr, dirname) -- iterate over the
    end
    return coroutine.wrap(iter), iter_dir
 end
+-- This will probably get exposed in the api if that happens
+function project:files(inc_patt_arr, exc_patt_arr, dirname) -- iterate over the files in the project adhering to the provided patterns
```
euclidianAce, line 877: This should just get deleted as it's no longer used and there's not much point to having an iterator that just throws away values from another since they can easily be ignored.

tl
```diff
@@ -538,7 +678,9 @@ if cmd == "run" then
    end
 
    tl.loader()
-   return chunk((unpack or table.unpack)(arg))
+   local res = chunk((unpack or table.unpack)(arg))
```
euclidianAce, line 681: Now that we're doing things after running the script, it might be worth it to pcall the script or wrap it in a coroutine just to guarantee the generated stuff gets cleaned up

tl
```diff
@@ -863,14 +970,18 @@ if cmd == "build" then
...
       --TODO: make this better
       local valid = true
       if not (path:match("%.tl$") and not path:match("%.d%.tl$")) then
          valid = false
       end
       if valid then
-         project.source_file_map[path] = path:gsub("%.tl$", ".lua")
+         local work_on = path:gsub("%.tl$", ".lua")
+         if is_build then
+            work_on = work_on:gsub(build_path,"")
```
euclidianAce, line 982: I think you want `remove_leading_path(build_path, work_on)` here as this `gsub` turns the path into an absolute one `foo/bar` -> `/bar`

tl
```diff
@@ -391,6 +411,109 @@ for _, include in ipairs(tlconfig["include_dir"]) do
...
+
+local build_path
+
+local function cleanup_generated_files()
+   if build_path then
+      delete_dir(build_path)
+   end
+end
+if tlconfig["build_file"] then
+   local tmp_dir = os.getenv("TMP") or os.getenv("TMPDIR") or "/tmp" --fall back to /tmp, apparently some (mine?) computers don't have a variable set for it.
```
euclidianAce, line 511: Do we even need to do things in `/tmp/`? Can we not stay within the project file in `./generated_code`?

tl
```diff
@@ -498,6 +621,23 @@ local function type_check_and_load(filename, modules)
...
+end
+
+if script["gen_code"] then
+   --that the directory exposed to the script has the same name as the the folder it ends up at is kind of a hack (that I accidentally stumbled on)
+   --however, it is also a usefull hack as it provides a way to store the last build time without the 2 locations interfering.
+   --this is useful information so the result of the build.tl file can be cached, which can be benefical if the code generation takes a lot of time.
+   local full_path = path_concat(build_path, tlconfig["build_file_output_dir"] )
+   lfs.rmdir(full_path)
+   lfs.mkdir(full_path)
+   script["gen_code"](full_path)
```
euclidianAce, line 639: Again, should type check this manually even though we loaded a `.tl` file, we didn't inject any types into it. We have no idea if this is a function and should check for that. Then it should be `pcall`ed, since it is user code.

docs/build_file.md
```diff
@@ -0,0 +1,36 @@
...
+You can configure where the lua files will be saved by setting `build_file_output_dir` in `tlconfig.lua`. This uses the directory set by `build_dir` as a base. The default value is `generated_code`.
+
+## Use case
+
+As mentioned earlier, this file can be used to generate types without the need for `make` or other external build tools. A reason why you might is if your teal code consumes an API that has schemas available.
+
+Then you could simply add these schemas to your repo and have the `build.tl` file create types based on these schemas for you. That way you only need to grab a new version of the schemas if they change.
+
+Another use case could be when using a teal version of a library like [pgtyped](https://github.com/adelsz/pgtyped) where you normally need to run a command manually to generate the types and code. Now you can just stick that in the `build.tl` file and forget about it.
+
```
euclidianAce, line 33: I think a small example that actually does something would be helpful here

spec/cli/build_script_spec.lua
```diff
@@ -0,0 +1,100 @@
+local util = require("spec.util")
+
+describe("build.tl", function()
```
euclidianAce, line 3: Some tests that check that errors from the build file would be nice

euclidianAce (changes requested):
Overall pretty decent, but I have left some comments with my concerns/questions.

One minor style thing that I noticed is you don't put spaces after your commas and this makes your additions look out of place, idk what editor you use but a quick search and replace of commas followed by a non-space should be easy enough. (If you use vim it should be something like `:%s/,\(\S\)/, \1/g`)

tl
lenscas: I guess I can make a folder in the project directory where the generated teal files are placed, this would make it easier to cache the results. However I fear that this will look a lot uglier from the user side of things, as teal suddenly will generate a folder outside of whatever directory was chosen as the `build_dir` and drop files there.

I'm also not sure how that would interact if the user also has a directory with that name if not every .lua file in the corresponding folder is/can be generated. (I hope this sentence makes sense...)

docs/build_file.md
lenscas: so, just have it create a simple teal file that returns a string or something?

lenscas:
ok, so the exact output that lua5.1 and luajit gives on syntax errors in the build.tl file is different than in the other versions.

It still contains all the information, so the output isn't actually wrong, just not exactly the same.

hishamhm:
> It still contains all the information, so the output isn't actually wrong, just not exactly the same.

I haven't looked at the test itself, but from the CI output, it looks like it's changing the output order of stderr and stdout? One way around this is to test for substrings of the output with `assert.match`, and not testing the whole output (one gotcha though is that `assert.match` matches a Lua pattern so characters like `-` have special meaning unless you force plain-match with `assert.match("this-thing-exists-inside", var, 1, true)` like `string.find`).

tl
```diff
@@ -391,6 +411,109 @@ for _, include in ipairs(tlconfig["include_dir"]) do
...
+local build_path
+
+local function cleanup_generated_files()
+   if build_path then
+      delete_dir(build_path)
+   end
+end
+if tlconfig["build_file"] then
+   local tmp_dir = os.getenv("TMP") or os.getenv("TMPDIR") or "/tmp" --fall back to /tmp, apparently some (mine?) computers don't have a variable set for it.
+   build_path = path_concat(tmp_dir,"tl_temp_out" .. os.time(os.date("!*t")))
```
hishamhm, line 512: > os.time(os.date("!*t"))

super minor thing because I happened to look at this bit of the code because of the comment above: I think `os.time()` with no arguments does the same?

tl
hishamhm: Perhaps something like `build_dir .. `/tl_build_generated` instead? This is an actually temporary directory which gets created then deleted, right?

tl
```diff
@@ -538,7 +693,10 @@ if cmd == "run" then
    end
 
    tl.loader()
-   return chunk((unpack or table.unpack)(arg))
+   local success,res = pcall(chunk,(unpack or table.unpack)(arg))
+   cleanup_generated_files()
```
hishamhm, line 697: Now I noticed one thing... is this PR running `build.tl` for _every_ possible `tl` command? I don't think we want to run it for every use of `tl run`. `tl run` is not intended to mean "run this project" as in `cargo run`, but instead it's more like "run this file" as in the Lua interpreter. 

Before looking at this PR, I assumed that only `tl build` would run the build file.

lenscas: check needs to run it as well, as otherwise it can't type check code depending on the generated files.

I can change it so that it doesn't run on the run command though if that is preferred

tl
lenscas: That could work as well, and you are correct in that it is a directory that gets created and later deleted.

However, why not use the tmp folder for temporary files? Is there something that I'm missing that makes not using the tmp folder preferred?

Also, eventually it is a good idea to only rerun the build file if it needs to which means that the user would see the `tl_build_generated` folder if we stick it in `build_dir` or anywhere else in the project, which doesn't sound nice if you ask me.

Unless we want to claim a folder for any build artifacts so that the files needed for https://github.com/teal-language/tl/issues/302 can be stored there as well?

In that case, I would recommend to make this folder changeable in tlconfig.lua and emit a warning (for now) if a build script is detected but this value is not set to make users aware of the fact that teal will "pollute" their project with files for internal use only. Once #302 is implemented this warning can either get lifted all together or get always emitted.

tl
lenscas: it could be, I searched how to get a unix timestamp (as I forgot) and this answer showed up first but now that you mention it, I am pretty sure as well that just `os.time()` is good enough as well.

lenscas: just checked it

> The time function, when called without arguments, returns the current date and time, coded as a number. (In most systems, that number is the number of seconds since some epoch.) 

https://www.lua.org/pil/22.1.html

I think the keyword is "most" in this case. However, as it is only used to reduce the chance of collisions (Especially with a previous run that couldn't delete the folder for one reason or another) I think that using `os.time` without arguments is still fine, for as long as those other ways have at least second precision.

hishamhm: > https://www.lua.org/pil/22.1.html

I think the above link is correct, but beware in general that lua.org/pil refers to Lua 5.0, which is out of date; usually only 5.1+ matter.

Given the [latest docs](https://www.lua.org/manual/5.4/manual.html#pdf-os.time) (which haven't changed much since 5.1), I think the outputs of `os.time` are just as system-dependent with or without arguments.

tl
hishamhm: Ermmm, `tl check` and `tl gen` also run file-by-file... it would be annoying to run it multiple times for different files and have them run auxiliary arbitrary scripts multiple times. `tl build` is already the command that means "do everything that needs to be done".

Are the results of `build.tl` preserved after running it once?

If so, Given a `build.tl`, I think a sensible procedure would be to expect users to run `tl build` once on a project, where it would build everything that needs to be pre-built, and then they would be able to run `tl check`, `tl run` and friends without regenerating everything.

In this first PR, I think we could do without a check that `tl build` was actually run first when needed and rely on the user to call things in the right order. As a future improvement, we could have a lock file of sorts in the project dir to detect that `build.tl` was already executed. How does that sound?

Side question: is the writer of `build.tl` expected to write it so it is idempotent?

lenscas: I think the various editor plugins rely on `tl check` though? Which is why I think that `tl check` should execute the build script if one is found.

> Are the results of build.tl preserved after running it once?

Not yet, that was planned to happen in another PR.

> Side question: is the writer of build.tl expected to write it so it is idempotent?

Short answer: Yes.

Longer explanation: Once caching works, the user is able to list a set of patterns where if a file changed that matches these patterns will invalidate the cache and thus trigger the script to rerun. If no files that match these patterns changed since last run (nor the build script itself) then teal will not rerun the script and reuse the previous files.



I can look into making caching happen in this PR, though before I do that I would like to have the point above resolved (The one about using the tmp folder vs a folder in the project).

hishamhm: > I would like to have the point above resolved (The one about using the tmp folder vs a folder in the project).

If we want to eventually have caching, then it should be a folder in the project, because folders under /tmp can end up being auto-cleaned-up by system services.

hishamhm: > I think the various editor plugins rely on tl check though? 

Yes, but editor plugins which launch compilers doing typecheck-as-you-type don't normally run multi-step build procedures generating files etc, right? For example, if I'm editing a C program in a `cmake` based project, I don't expect it to run the whole build just because I opened a file to edit it.

> If we want to eventually have caching, then it should be a folder in the project, because folders under /tmp can end up being auto-cleaned-up by system services.

Also, I don't mind starting with things in /tmp as they are in this PR but then moving later to the project when caching is added, but I think that at this time `build.tl` should be triggered only by `tl build`.

lenscas: > Yes, but editor plugins which launch compilers doing typecheck-as-you-type don't normally run multi-step build procedures generating files etc, right? For example, if I'm editing a C program in a cmake based project, I don't expect it to run the whole build just because I opened a file to edit it.

If I open a rust project, then yes that is pretty much what happens. `rust-analyzer` runs `cargo check` which results in cargo to download the dependencies if needed and build everything (though it won't spit any actual binaries out as it stops before that phase). This does include running build files however.

> Also, I don't mind starting with things in /tmp as they are in this PR but then moving later to the project when caching is added, but I think that at this time build.tl should be triggered only by tl build.

I already started to move things out of `/tmp` and started work on the caching system. Right now it only checks if he build script changed before running again. I haven't pushed it yet as there are still some bugs left.

hishamhm: > If I open a rust project, then yes that is pretty much what happens. rust-analyzer runs cargo check which results in cargo to download the dependencies if needed and build everything (though it won't spit any actual binaries out as it stops before that phase). This does include running build files however.

One thing to note is that `cargo` is a full-fledged package manager. Right now `tl build` is the "odd one out" in the `tl` CLI because it does "package-manager-like things" you'd expect from `cargo` or `npm`, while things like `tl gen`, `tl run`, `tl check` are more analogous to `rustc` or `node`. 

I'm not opposed in principle to having `tl` have a broad scope (`tl build` is already in, after all), but I still want to have the "low-level commands" that operate on the files they are given as arguments. I'm not surprised if `tl build` does a lot of things (and someday even downloads dependencies and whatnot) but I would never expect `tl check file.tl` to do anything other than running the typechecker on `file.tl` and certainly not running some of the project's arbitrary code.

> I think the various editor plugins rely on tl check though? Which is why I think that tl check should execute the build script if one is found.

Most (all?) plugins were written before `tl build` existed. I'm inclined to think that what's needed is to make editor plugins aware of `tl build`, instead of adding build logic into `tl check`.

lenscas: what about being able to pass an argument to teal to also run the build script? And/or have an extra command that runs the build script if needed? In both cases `tl build` would of course still run it if needed as that one is meant to do everything to get the project up and running.

hishamhm: > what about being able to pass an argument to teal to also run the build script? And/or have an extra command that runs the build script if needed? In both cases tl build would of course still run it if needed as that one is meant to do everything to get the project up and running.

All of these options are fine by me! I'd just prefer the "low-level commands" that take on a `.tl` file as an argument (`check`, `gen`, `run`) to not do it automatically.

lenscas: decided to go with an argument.

hishamhm:
@lenscas just checking in on the status of this PR — is it ready for another round review or do you plan to add anything else?

@euclidianAce I see you have marked "changes requested" and various changes were made since — could you take another look, in case the PR is ready for re-review? Thanks!

lenscas:
I think so, only thing I didn't do is make a test for the caching behavior, not sure how to best do that.

hishamhm:
> a test for the caching behavior, not sure how to best do that.

depending on how deterministically the cache behaves, one possibility would be to:

1. run thing once
2. check presence of file in cache, store file timestamp
3. run thing again
4. check presence of file in cache, verify file timestamp is the same (i.e. second run didn't regenerate file in cache)

lenscas:
`lfs.attributes(path,"modification")` returns the time in seconds, so I'm not sure if it is precise enough. (As to my experience teal is generally rather fast, especially with simple projects.

hishamhm:
@lenscas ah, you're right. Can you add a `print` to the `build.tl` file being tested, so that you can verify the output of the command and check that the message from `build.tl` is printed in the first invocation but not the second?

lenscas:
Test is added

euclidianAce (approved):
Looking good. Just one thing as mentioned before, you don't put spaces after your commas some some bits just look out of place style-wise, I can handle this once its merged if you don't want to/aren't able to

lenscas:
I don't use Vim, so unless there is an easy way in vscode that I don't know about, I would have to go over the functions by hand. So, if you are able to do then that would be great :)

hishamhm:
@euclidianAce @lenscas all right, should I go ahead and merge this, and then we make any further tweaks?

lenscas:
> and then we make any further tweaks

After it is merged I will start with the next PR, which will focus on being able to do stuff after compiling (only for `tl build`, and able to stop the compile process in a nicer way than throwing an error. Once that PR lands, you should be able to run stuff like luapak on the compiled output without the need for `make` / another build tool or dedicated compiler integration.

I consider it a success if I can rewrite the `make` file that teal uses into the build script, note: ***can*** is the magic word, wether or not teal should switch is a separate discussion.

hishamhm:
@lenscas squash-merged! :tada:

PR #304, opened 2020-12-29
Rework tests to not use lfs.link + refactor temp file creation

euclidianAce:
- This was the easy solution that worked at first when `tl build` was
   implemented, but making a bunch of random links can cause problems,
   so now we just run the `tl` command given its full path and tell
   busted to add the current dir to the lua path so `tl` can find
   `tl.lua`
 - write_tmp_file will now generate its own name and now
   accepts a parameter for the extension of the file it generates, i.e.
   previously: `write_tmp_file(finally, name, content)`
          now: `write_tmp_file(finally, content, extension)`
 - Add "tl", and ".busted" to .editorconfig

### DIFF
--- a/.busted
+++ b/.busted
@@ -1,5 +1,6 @@
 return {
-  default = {
-    lpath = "./?.lua";
-  }
+   default = {
+      lpath = "./?.lua;"
+         .. (require("lfs").currentdir()) .. "/?.lua;";
+   }
 }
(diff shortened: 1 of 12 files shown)


hishamhm:
FWIW, this branch ran with no errors on my machine. CI runs are also clean, so LGTM!

hishamhm (approved):

PR #309, opened 2021-01-01
Make it so function calls in expression lists get adjusted to 1 result

euclidianAce:
No idea if this has caused any issues but I stumbled upon the following behavior:
```
local function f(): string, number
end

local a, b = f(), 1
print_type(a)
```
`a` is assigned the tuple `(string, number)`, and in the following case
```
local function f()
end
local a, b, c = 1, f(), 3
print_type(b)
```
`b` is an empty tuple `()`

Anyway, this fixes that and I'm not exactly sure how to write a test for this or even if it's needed

### DIFF
--- a/tl.tl
+++ b/tl.tl
@@ -5963,13 +5963,7 @@ tl.type_check = function(ast: Node, opts: TypeCheckOptions): {Error}, {Error}, T
       -- get all arguments except the last...
       local is_va = vals.is_va
       for i = 1, #vals - 1 do
-         if vals[i].typename == "tuple" then
-            -- result of function call in explist
-            -- adjust to 1 result
-            ret[i] = vals[i][1] or NIL
-         else
-            ret[i] = vals[i]
-         end
+         ret[i] = resolve_tuple(vals[i])
       end
 
       local last = vals[#vals]
--- a/tl.tl
+++ b/tl.tl
@@ -5963,7 +5963,13 @@ tl.type_check = function(ast: Node, opts: TypeCheckOptions): {Error}, {Error}, T
       -- get all arguments except the last...
       local is_va = vals.is_va
       for i = 1, #vals - 1 do
-         ret[i] = vals[i]
+         if vals[i].typename == "tuple" then
+            -- result of function call in explist
+            -- adjust to 1 result
+            ret[i] = vals[i][1] or NIL
+         else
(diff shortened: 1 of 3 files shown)


lenscas:
> I'm not exactly sure how to write a test for this
Maybe call a function with `a` that only works if it is the correct type? Same with `b` in the second case?

tl.tl
```diff
@@ -5963,7 +5963,13 @@ tl.type_check = function(ast: Node, opts: TypeCheckOptions): {Error}, {Error}, T
...
       local is_va = vals.is_va
       for i = 1, #vals - 1 do
-         ret[i] = vals[i]
+         if vals[i].typename == "tuple" then
+            -- result of function call in explist
+            -- adjust to 1 result
+            ret[i] = vals[i][1] or NIL
+         else
+            ret[i] = vals[i]
+         end
```
hishamhm, line 5972:
```suggestion
         ret[i] = resolve_tuple(vals[i])
```

hishamhm:
> Anyway, this fixes that and I'm not exactly sure how to write a test for this or even if it's needed

It is a valid bug, good catch! Here's a test case that demonstrates it:

```lua
local function f(): string, number
end

local a, b, c = 1, f(), 3

local x, y = b
```

On `master` this passes; with the fix it produces a type error (correctly) because `y` does not have a type assigned to it.

hishamhm:
( @darrenjennings a bit offtopic to this PR but: the playground for this PR is failing as well (I tried re-running Actions, same result) — what it has in common with the previous failure is that this by @euclidianAce as well, and the one thing I noticed that's different from other examples that worked is that these PRs use non-master branches; however, the tl.lua URL seems to work; I tried concatenating the URL by hand based on what the action seems to do and  https://raw.githubusercontent.com/euclidianAce/tl/exprlist-tuple-adjustment/tl.lua seems to work — any idea? btw, I don't know if it helps, but I noticed that if I take the PR's SHA and use it with `teal-language/tl`, it _also_ works and returns the same file: https://raw.githubusercontent.com/teal-language/tl/f17700881d5445d1ac2ff57b725ff224e51abf52/tl.lua )

darrenjennings:
@hishamhm I found the issue and filed it here so we can implement a workaround: https://github.com/teal-language/tl/issues/314

PR #312, opened 2021-01-03
Implement warning tags

euclidianAce:
Tag each warning so that they can be suppressed or promoted to errors
via user config + implement the cli and config options to do so

adds the `--wdisable` and `--werror` cli flags and the `disable_warnings` and `warning_error` config entries


euclidianAce:
Apparently some of the tests weren't able to find the `tl` module? I think the tests that are failing have to do with the tests that chdir into tmp not being able to find `tl.lua`? Taking a quick glance at the travis file, would this be fixed by directly executing `luarocks install tl-dev-1.rockspec` so that it puts `tl.lua` into the lua path? or `luarocks make` might be the correct one?

hishamhm:
> Taking a quick glance at the travis file, would this be fixed by directly executing luarocks install tl-dev-1.rockspec so that it puts tl.lua into the lua path? or luarocks make might be the correct one?

That would be `luarocks make`, but if the tests start relying on the systemwide `tl.lua`, this might cause trouble for those of us who have an installed version of `tl` but want to develop on a different one. 

Still, this is an odd error to have, because I thought your `lfs.currentdir()` trick in `.busted` took care of setting the absolute dir in the `package.path` — doesn't `.busted` get processed only once at startup? Odd.

hishamhm:
> Still, this is an odd error to have, because I thought your lfs.currentdir() trick in .busted took care of setting the absolute dir in the package.path — doesn't .busted get processed only once at startup? Odd.

Ah, rereading the test, the busted configuration probably sets it for the Lua VM that running busted, but the tests in question os.execute a separate tl process. Perhaps our util that runs things could prepend something like `"LUA_PATH='" .. package.path .. "' LUA_PATH_5_2='" .. package.path .. "'`...

euclidianAce:
Okay, that worked. I thought I did that at some point before but apparently not :laughing:.

spec/util.lua
```diff
@@ -134,6 +135,11 @@ function util.write_tmp_file(finally, content, ext)
    assert(type(finally) == "function")
    assert(type(content) == "string")
 
+   -- TODO: since busted runs tests concurrently, the following race condition happens occasionally
+   --    test a: starts, touches /tmp/teal_tmp_file1.tl
+   --    test b: finishes, deleting /tmp/teal_tmp_file1.tl
+   --    test a: attempts to read from /tmp/teal_tmp_file1.tl, fails
+   -- solution: use rng?, a simple uuid impl? set an environment var to increment?
```
hishamhm, line 142: yes, I think adding a `math.random(999999)` to the filename would suffice!

spec/util.lua
```diff
@@ -111,6 +111,7 @@ function util.tl_cmd(name, ...)
    assert(valid_commands[name], "not a valid command: tl " .. tostring(name))
 
    local cmd = {
+      string.format("LUA_PATH=%q", package.path),
```
hishamhm, line 114: this is sufficient for CI because the environment is clean, but it might be safest to add all versioned variants (`LUA_PATH_5_2`, `LUA_PATH_5_3`, `LUA_PATH_5_4`) as well

hishamhm:
left two suggestions, but overall LGTM!!

hishamhm:
Merged! :rocket:

PR #316, opened 2021-01-04
fix: adjust playground url

darrenjennings:
- remove username to avoid case sensitivity

@hishamhm I originally only used the username so that the [url looks good](https://github.com/teal-language/tl/blob/master/.github/workflows/playground.yml#L39) and and appending the PR `id` for uniqueness. They don't have any bearing on the tl [source links](https://github.com/teal-language/tl/blob/master/.github/workflows/playground.yml#L33). After looking at it, I think maybe we can just use the `pull_request.number` instead. So playground urls would change like so:

e.g. https://github.com/teal-language/tl/pull/315

```diff
-https://hishamhm-547967880--teal-playground.netlify.app
+https://315--teal-playground.netlify.app
```

Let me know if that works for you, unless having the username has benefits, then we can lower it in the script.

### DIFF
--- a/.github/workflows/playground.yml
+++ b/.github/workflows/playground.yml
@@ -36,7 +36,7 @@ jobs:
       run: |
         yarn global add netlify-cli
         cd ${{ github.workspace }}/teal-playground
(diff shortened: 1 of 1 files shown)


hishamhm:
I think that works too! :+1: Your call!

Issue #317, opened 2021-01-04
A few questions and comments


FourierTransformer:
So, I was playing around with tl for ftcsv (in [this branch](https://github.com/FourierTransformer/ftcsv/tree/convert-to-tl)) and it is really cool, and pretty easy to use. I'm glad to see a minimalistic type system for lua! Congrats to the team behind it! I even found a bug because of it!

1) I did have a couple of questions about using tl. I have this basic file loading snippet, and depending on the ftcsv options will either read in the whole file (with `*alll`) or just a certain number of bytes. So I went with the type definition of `string | number` but noticed I still needed to typecheck to use it in `file:read()` even though `read()` can take in either. Am I doing something wrong with the union or is this the way it is?
```lua
local function loadFile(textFile: string, amount: string | number)
    local file = io.open(textFile, "r")
    if not file then error("ftcsv: File not found at " .. textFile) end
    local lines: string
    if amount is string then
        file:read(amount)
    else
        file:read(amount)
    end
    if amount == "*all" then
        file:close()
    end
    return lines, file
end
```

2) The other thing I ran into was using `{{string:any}}`. When trying to encode a csv from a lua table that looks something like this: `a= {{b="hey", c=true, d=43}, {b="yo", c=false, d=42}}`. In some of the encoder code, I was trying to use `{{string:any}}` but ran into the error below. I worked around it by creating a `local type CSVRow = {string: any}` and switching form `{{string:any}}` to `{CSVRow}`. I'm mostly curious if that's the intended way to fix that issue and if I contributed types for ftcsv in the type definitions project would people need to cast their tables as `CSVRow` to make things work?
```
5 errors:
encoder.lua:90:5: in assignment: got {{string : <any type>}}, expected {{string : <any type>}}
encoder.lua:152:43: argument 1: got {{string : <any type>}}, expected {{string : <any type>}}
encoder.lua:154:30: argument 2: got {{string : <any type>}}, expected {{string : <any type>}}
encoder.lua:163:49: argument 1: got {{string : <any type>}}, expected {{string : <any type>}}
encoder.lua:165:37: argument 1: got {{string : <any type>}}, expected {{string : <any type>}}
```

3) Can type definitions be auto-generatred from tl files? For my stuff it's not a lot to write, but would be cool for other stuff!

I can create separate issues if needed. So far in my time with tl, it's been pretty amazing!

hishamhm:
> I even found a bug because of it!

[...]

PR #319, opened 2021-01-04
fix: maps are invariant on keys and values

### LINKED ISSUE #318: odd "got 'type', expected {}" in map declaration
example:
```
local map: {boolean:{number}} = {
	[true] = {},
}
print_type(map)
```
```
$ tl check blah.tl
({boolean : {number}})
2 errors:
blah.tl:2:22: in local declaration: map: got {number}, expected {}
blah.tl:5:1: unknown variable: print_type
```
This only seems to happen when the keys are of a certain type. For example, `number` keys are just fine
```
local map: {number:{number}} = {
	[3] = {},
}
```
type checks just fine

Additionally, it seems like some `same_type`/`is_a` arguments are in the wrong order somewhere as the intent here seems like it should be `got {}, expected {number}`

hishamhm:
Fixes #318.

### DIFF
--- a/spec/assignment/to_array_of_maps_spec.lua
+++ b/spec/assignment/to_array_of_maps_spec.lua
@@ -13,4 +13,10 @@ describe("assignment to array of maps", function()
          },
       }
    ]])
+
+   it("resolves subtypes correctly for empty tables (#318)", util.check [[
+      local map: {boolean:{number}} = {
+         [true] = {},
+      }
+   ]])
 end)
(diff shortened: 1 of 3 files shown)


FourierTransformer:
Can confirm this fixed the one issue described in #317 !

hishamhm:
@euclidianAce Looks like this fixes #318 as well, any concerns on merging this one?

euclidianAce:
@hishamhm Looks good to me :+1:

hishamhm:
Thank you for the feedback! Merged!

PR #320, opened 2021-01-04
stdlib: prefer unions over poly whenever possible

hishamhm:
See #317.


FourierTransformer:
Cool, this also worked out! Also, the teal-playground is really cool - thanks for the quick turnaround!

hishamhm:
thanks for reporting and testing!

PR #325, opened 2021-01-08
fix type comparison of tuple elements

hishamhm:
As a side effect, allows `{T, nil}` to be used as an idiom for an
effectively unary tuple (since `{T}` means array of T).


hishamhm:
@euclidianAce any concerns come to mind about this one? it's a small change to the tuple code.

euclidianAce:
Looks good to me. As far as unary tuples go, I did think about adding a shorthand syntax for tuples of a single type/fixed length arrays to just give a type + a length. Perhaps sometime in the future that could be implemented.

PR #333, opened 2021-01-13
fix invariant comparison between {} and nominals

### LINKED ISSUE #332: tl failed to resolve map value type with some format key
source code is

```
local keystr = "aaaa"

local record user_login_count_t
end

local data:{string: user_login_count_t} = {
    key={},
    [keystr]={}
}

for k, v in pairs(data) do
    print(k, v)
end
```

the error message is

```
local declaration: data: got {}, expected user_login_count_t
```

hishamhm:
This is a regression detected when maps keys and values switched from
covariant to invariant.

Fixes #332.

### DIFF
--- a/spec/assignment/to_map_spec.lua
+++ b/spec/assignment/to_map_spec.lua
@@ -20,4 +20,20 @@ describe("assignment to maps", function()
          world = "south",
       }
    ]])
+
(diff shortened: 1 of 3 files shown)


virusdefender:
it works now ~

hishamhm:
@virusdefender thanks for confirming!

PR #338, opened 2021-01-15
Fill out debug std lib

euclidianAce:


tl
```diff
@@ -323,7 +323,7 @@ local function report_errors(category, errors)
       printerr("========================================")
       printerr(n .. " " .. category .. (n ~= 1 and "s" or "") .. ":")
       for _, err in ipairs(errors) do
-         printerr(err.filename .. ":" .. err.y .. ":" .. err.x .. ": " .. (err.msg or ""))
+         printerr(err.filename .. ":" .. (err.y or "?") .. ":" .. (err.x or "?") .. ": " .. (err.msg or ""))
```
hishamhm, line 326: Oh, have you bumped into an error without a location? Perhaps we should be dealing with that by fixing the error instead?

euclidianAce: It happened a couple times when trying to add the `debug.Hook` and `debug.HookEvent` types since I accidentally added them as nominal type rather than a typetype, it should work without that now

hishamhm: Got it. I'd rather let that crash (which would make it more likely we'll get a bug report) than produce a `:?:?` in case some actual error is generated without a location (which would more likely cause the user to shrug and move on)

euclidianAce: Yeah, makes sense

PR #347, opened 2021-01-20
Fix casting empty expressions + never error with just "syntax error"

### LINKED ISSUE #345: Compiler crash when use parenthesis in return
Example
```lua
function foo(): number
    return (1, 1) as number
end
```
Report
```
/usr/bin/lua5.4: /home/testuser/.luarocks/share/lua/5.4/tl.lua:1723: attempt to index a nil value (local 'e1')
stack traceback:
        /home/testuser/.luarocks/share/lua/5.4/tl.lua:1723: in upvalue 'P'
        /home/testuser/.luarocks/share/lua/5.4/tl.lua:1759: in local 'parse_item'
        /home/testuser/.luarocks/share/lua/5.4/tl.lua:1279: in upvalue 'parse_trying_list'
        /home/testuser/.luarocks/share/lua/5.4/tl.lua:2306: in function </home/testuser/.luarocks/share/lua/5.4/tl.lua:2301>
        (...tail calls...)
        /home/testuser/.luarocks/share/lua/5.4/tl.lua:2479: in upvalue 'parse_statements'
        /home/testuser/.luarocks/share/lua/5.4/tl.lua:1507: in upvalue 'parse_function_args_rets_body'
        /home/testuser/.luarocks/share/lua/5.4/tl.lua:1898: in function </home/testuser/.luarocks/share/lua/5.4/tl.lua:1867>
        (...tail calls...)
        /home/testuser/.luarocks/share/lua/5.4/tl.lua:2479: in upvalue 'parse_statements'
        /home/testuser/.luarocks/share/lua/5.4/tl.lua:2522: in function 'tl.parse_program'
        /home/testuser/.luarocks/share/lua/5.4/tl.lua:7671: in function 'tl.process_string'
        /home/testuser/.luarocks/share/lua/5.4/tl.lua:7634: in function 'tl.process'
        ...user/.luarocks/lib/luarocks/rocks-5.4/tl/0.10.1-1/bin/tl:907: in main chunk
```

euclidianAce:
Previously expressions like `() as string` and `1 as` would crash the
compiler, these now produce syntax errors instead.

Additionally, any call to `fail` now must have an associated error
message

Fixes #345

### DIFF
--- a/tl.tl
+++ b/tl.tl
@@ -1105,7 +1105,7 @@ local function fail(ps: ParseState, i: number, msg: string): number
       table.insert(ps.errs, { y = eof.y, x = eof.x, msg = msg or "unexpected end of file" })
       return #ps.tokens
    end
-   table.insert(ps.errs, { y = ps.tokens[i].y, x = ps.tokens[i].x, msg = msg or "syntax error" })
+   table.insert(ps.errs, { y = ps.tokens[i].y, x = ps.tokens[i].x, msg = assert(msg, "syntax error, but no error message provided") })
    return math.min(#ps.tokens, i + 1)
 end
 
@@ -1186,7 +1186,7 @@ end
 local function parse_table_item(ps: ParseState, i: number, n: number): number, Node, number
    local node = new_node(ps.tokens, i, "table_item")
    if ps.tokens[i].kind == "$EOF$" then
-      return fail(ps, i)
+      return fail(ps, i, "unexpected eof")
    end
 
    if ps.tokens[i].tk == "[" then
@@ -1239,7 +1239,7 @@ local function parse_table_item(ps: ParseState, i: number, n: number): number, N
    node.key.tk = tostring(n)
    i, node.value = parse_expression(ps, i)
    if not node.value then
-      return fail(ps, i)
+      return fail(ps, i, "expected an expression")
    end
    return i, node, n + 1
 end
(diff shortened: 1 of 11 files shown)


hishamhm:
There are cool improvements here, but I'd like to point out that some generic syntax error messages were deliberate.

Sometimes the messages make sense if you're reading it from the perspective of the parser/grammar ("expect an assignment or function call") but from a user perspective this message happens when a user makes a typo and write something like `iff x < 2 then`. It doesn't really make sense to say "expected either an assignment or function call" there, even though it's technically correct _from a grammar point of view_.

I avoided just putting `"expected foo"` for every error happening in a rule `parse_foo`, and tried to make sure that messages only triggered in situations where they would make sense (e.g. there's only one valid token that could happen there). For example, I put "expected expression" in the context of operator errors.

IMO whenever the parser has a hard time pinpointing the intent of the source code in an error, it's better to just point to the exact line-column location and let the user spot what's wrong, which is usually _faster_ than having them try to decode an error message that is technically correct but doesn't make much sense (I remember many times I've had typos in C code and gcc reported me errors including a long list that went like `expected one of: (, {, __asm__, sizeof, ...` and I was like "huh? `__asm__`??"). If you output too much useless stuff in error messages, users stop reading them.

I'll go through the errors and do some tweaks prior to merging this manually. Thanks for the initiative! Improving error messages is always a good thing!

tl.tl
```diff
@@ -1685,6 +1685,10 @@ do
          i, e1 = parse_literal(ps, i)
       end
 
+      if not e1 then
+         fail(ps, i, "expected an expression")
+         return i
+      end
```
hishamhm, line 1691: if `e1` doesn't exist, another error must have been already through in all paths leading here, right? I've been trying to not output repeated errors (a recent commit got rid of a few recently)

hishamhm:
I took the liberty to change it around a bit: d119ddb — I restored some generic `syntax error`s where I felt it made more sense intentionally (added a paragraph in the commit message with the rationale) and also added error messages for malformed strings and numbers. Thank you for the nudge!

hishamhm:
Closing since it was merged manually!

PR #356, opened 2021-01-26
Disable JIT and GC when compiling

hishamhm:
The compiler is a short-lived process, so JIT (on LuaJIT) and the GC
add unnecessary overhead.

Since `tl run` executes arbitrary code from the user, we reenable
the JIT and GC prior to running the user's code.

### DIFF
--- a/spec/util.lua
+++ b/spec/util.lua
@@ -1,5 +1,10 @@
 local util = {}
 
+if jit then
+   jit.off()
+end
+collectgarbage("stop")
+
 local tl = require("tl")
 local assert = require("luassert")
 local lfs = require("lfs")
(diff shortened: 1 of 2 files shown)


lenscas:
Maybe it is a good idea to give users a way to opt out of turning off the GC? It would suck if someone is unable to use teal simply because the compiler uses too much memory.

hishamhm:
> It would suck if someone is unable to use teal simply because the compiler uses too much memory.

In my measurements, there is no significant difference in memory consumption (it's a bunch of small objects). But I made some changes to stay on the safe side. It will run GC on larger projects; managed to avoid a flag.

lenscas:
> In my measurements, there is no significant difference in memory consumption (it's a bunch of small objects)

That doesn't surprise me, but I feel that things like disabling the GC is fine until it suddenly isn't. Which is why I suggested a way to turn it back on again before someone reached that point :)

PR #358, opened 2021-01-26
Build script: Run code afterwards

lenscas:
Lots of things that need to happen still, but it is at least a start. I think I am going to split it up as multiple pr's to reduce the amount of code changes in 1 pr.

First PR: The absolute basics.

- [x] move the generated code to the correct folder. 
- [x] get a temporary folder in a way that windows likes as well.
- [x] clean up after it is done ?

Second PR small improvements:
- [ ] give more information to gen_code than just the folder for generated .tl files. (Location for generated .lua files and maybe the project folder itself?)
- [x] allow the build script to run code after the compile step. (Probably only for `tl build`)
- [x] have a good way to tell the build script to stop compiling, as things are broken. (Just return a boolean + error message?.)

Third PR cache the results: 

- [x] keep track of when the build script ran 
- [ ] clean up old artifacts when build script needs to rerun.

### DIFF
--- a/tl
+++ b/tl
@@ -668,7 +668,7 @@ end
 local internal_output
 if check_should_run_build_script() and tlconfig["internal_compiler_output"] then
 
-   internal_output = path_concat(lfs.currentdir(),tlconfig["internal_compiler_output"])
+   internal_output = path_concat(lfs.currentdir(), tlconfig["internal_compiler_output"])
    local mode = lfs.attributes(internal_output,"mode")
    if not mode then
       local parts = ""
@@ -691,15 +691,14 @@ end
 --------------------------------------------------------------------
 
 if check_should_run_build_script() and tlconfig["build_file"] then
-   build_path = path_concat(internal_output,"build_script_output")
+   build_path = path_concat(internal_output, "build_script_output")
    lfs.mkdir(build_path)
    prepend_to_path(build_path)
 end
 
-
+local script
 if check_should_run_build_script() and tlconfig.build_file then
-   local script = {}
-   local chunk = type_check_and_load(tlconfig.build_file,modules)
+   local chunk = type_check_and_load(tlconfig.build_file, modules)
    local success,res = pcall(chunk)
    if success then
       script = res
@@ -720,9 +719,13 @@ if check_should_run_build_script() and tlconfig.build_file then
          local full_path = path_concat(build_path, tlconfig["build_file_output_dir"] )
          lfs.rmdir(full_path)
          lfs.mkdir(full_path)
-         local success,message = pcall(gen_code,full_path)
+         local success, messageOrShouldContinue, message = pcall(gen_code,full_path)
          if not success then
-            die("Something has gone wrong while executing the \"gen_code\" part of the build file. Error : ".. tostring(message))
+            die("Something has gone wrong while executing the \"gen_code\" part of the build file. Error : ".. tostring(messageOrShouldContinue))
+         end
+         if messageOrShouldContinue ~= nil and not messageOrShouldContinue then
+            local reasonMessage = (message and "Reason:\n" .. message) or ""
+            die("Build interupted by build script." .. reasonMessage)
          end
          local file = io.open(time_keeper_path, "w")
          file:write(last_edit_time)
@@ -986,7 +989,7 @@ project.dir, project.paths, project.emptyref = traverse(lfs.currentdir())
 local build_ref = {}
 project.generatedref = build_ref
 if build_path then
-   local build_dir, build_paths, emptyref = traverse(build_path,project.emptyref,true,build_ref)
+   local build_dir, build_paths, emptyref = traverse(build_path, project.emptyref, true, build_ref)
    for k,v in pairs(build_dir) do
       project.dir[k] = v
    end
@@ -1119,6 +1122,17 @@ for i, files in ipairs(sorted_source_file_arr) do
       write_out(result, output_file)
    else
       exit = 1
+   end
+end
+if script then
+   if script["after"] and type(script["after"]) == "function" then
+      local path = path_concat(lfs.currentdir(), tlconfig["build_dir"])
+      local success,message = pcall(script["after"], path)
+      if not success then
+         die("Something has gone wrong while executing the \"after\" part of the build file. Error : ".. tostring(message))
+      end
+   else
+      die("the key \"after\" exists in the build file, but it is not a function. Value : " .. tostring(script["after"]))
    end
 end
 os.exit(exit)


lenscas:
Tests are coming soon. But first:

Right now, the function that runs after compiling is called "after". However, I'm not pleased with that name. Suggestions are welcome.

should `gen_code` get a series of paths? So, the first is the path where it should put teal files, then the path of the project itself and lastly the path where every compiled file ends up? (in the case of `tl build`) Or should those other paths be put in a table?

The table makes it both easier to add more information later but it also makes it more annoying to use compared to just 3 string parameters unless a type for it gets injected.

euclidianAce:
I'm gonna just bring up some talking points that I've thought about as I'm currently working on a pr for lazy rebuilds and doing some decently sized refactors/reorganization to the `tl` script. (So maybe hold off on any big features here until that lands :) )

Personally, when I work with a compiled language, it's either
 - C: which has no real (built-in) build system, but `Makefile`s and whatnot. or
 - Zig: which has a `build.zig` file that is more analogous to a cargo.toml or makefile than `build.rs`

So I'm not super familiar with Rust and it's `build.rs` file, which is what this feature seems to be emulating. My understanding of it (and Rust's build system) is that you have a `cargo.toml` which is like a rockspec/gemspec/package.json/whatever that describes versioning and dependencies. And then you can also have a `build.rs` which, (as this feature is trying to emulate) can run some callbacks pre and post build.

But when I think 'build script', I think of `build.zig` which combines the functionality of describing a packages version, dependencies, etc., and describing build steps. So I'll just be blunt here, I do not see the point of having a separate `build.tl` file for this. I'd say either abandon `tlconfig.lua` and have `build.tl` be the sentinel that marks a teal project, or move these features into `tlconfig.lua` and get rid of the separate file.

and upon some further reading: a decent amount of what a rust build script does can already be done by `tlconfig.lua`. Adding to the path (`include_dir`), passing flags to the compiler (we really only have `gen_target`and `gen_compat` really), etc. So I'm just curious about why this should be a separate file other than similarity to Rust.

If there's some functionality that I'm blantantly missing or some functionality planned that justifies this being a separate file I'd love to hear it

lenscas:
> some functionality planned that justifies this being a separate file

If it becomes part of tlconfig.lua, then you don't have type checking in this file. For me, that is enough reason for wanting it separately from tlconfig.lua but perhaps I am in the minority in that?

> If there's some functionality that I'm blantantly missing

build.rs is also used to generate code/types. Admittedly, it is less often used for that but this is because Rust has access to macro's that offer more flexibility to the user.

Still, I think that stuff like [pgtyped](https://github.com/adelsz/pgtyped) can (and should) come to teal and the build.tl file could help with streamlining using it as it will execute the correct commands (and in the correct order) for you after you made it. If dependencies can also have a build.tl file you won't even need to worry about making one, but that would probably require a different package manager which isn't something I am in favor of.

As for the use of "after", lua is used in a lot of places and expect code in a lot of formats. 

* If I remember correctly, love2D needs it all in a zip like file, so this feature can be used to automatically zip it. 

* There are probably places which expect it as a single file, so why not have luapak automatically run?

* My cardgame expects the card to be in json format containing lua code. Right now I have a separate teal program that generates these json files based on the teal files for the cards. With "after" I can do this automatically as well.

euclidianAce:
> If I remember correctly, love2D needs it all in a zip like file, so this feature can be used to automatically zip it.
> There are probably places which expect it as a single file, so why not have luapak automatically run?

iirc love2D can package stuff into an exe for release but it isn't necessary just to run/test it. These both sound more like 'release' builds rather than something I'd want to run on every `tl build`, (Having a `tl build release`, `tl build test`, etc and just build subcommands in general do sound nice tho).

> If it becomes part of tlconfig.lua, then you don't have type checking in this file.

Honestly, I'm okay with changing it to `tlconfig.tl` if we start putting callbacks to run in it, but it may give the wrong impression to users expecting type checking where we can't do it yet. (at least without hacks, like verifying the return type of the table returned by tlconfig)

lenscas:
I just checked. Love2D can indeed run from a folder. As for the single file one, it probably depends what you are using it for.

> (Having a tl build release, tl build test, etc and just build subcommands in general do sound nice tho

They do indeed, though a `tl build release` may get people to think that it will do optimizations, rather than just running a (different) script after compiling. Then again, its a good thing to have if teal starts doing optimizations

> Honestly, I'm okay with changing it to tlconfig.tl if we start putting callbacks to run in it, but it may give the wrong impression to users expecting type checking where we can't do it yet. (at least without hacks, like verifying the return type of the table returned by tlconfig)

Well, `tl` already does some basic type checking on the config file, so depending on what users exactly expect that is good enough? Worst case we can just make a .d.tl file that you can require if you want to. 

Maybe its a good idea to start a discussion about moving tlconfig.lua to a teal file?

hishamhm:
I don't have a lot to add to the specifics but I like seeing this discussion happening.

So bear with me, I was working as a Product Manager in my dayjob until recently (and PM work is often implicit/invisible in OSS projects), so let me put my PM hat back on and let's look at things from a high-level perspective. :)

I think the central question being raised here is what is the scope we see for the `tl` CLI tool. Everyone comes in with different expectations based on the environments they're familiar with. @euclidianAce summarized some of the questions well.

I'll be upfront and say that I _don't_ see myself as the most qualified person to decide on the UX of the CLI experience, mostly because I've been tainted by too much manual wrangling in the Lua ecosystem, and there have been advances in other environments which I haven't had significant hands-on experience with, even though I've tried to stay reasonably aware of what's going on. 

I like that the `go` experience allows for a pretty seamless newcomer experience (it's a positive sign that I managed to successfully use it at work _without_ becoming an expert on it!). I see that `cargo` in Rust is well loved by its community, and that approach essentially means making the package manager to be the main UI between the programmer and the language, not the compiler (I've also managed to successfully integrate Rust code in a project without ever touching `rustc` or being even aware of its existence!). Then there's `node` and `npm` (the largest package ecosystem around), Zig (which I haven't used personally, but from @euclidianAce's description it's essentially "let's use the language itself as its build tool", and I'm aware they do a good job in providing a self-contained package with everything you need to get up and running, LLVM and the kitchensink included), etc.

I've worked on LuaRocks for a long time (and before that, on GoboLinux, so LuaRocks wasn't even the first package manager I wrote). Package management can be a pretty annoying problem to work on: the corner cases pile up fast if you try to address everyone's use-cases, and given the general lack of consensus, the Lua world is full of incompatible sub-worlds, which make the situation even worse. The success of LuaRocks at organizing the messy Lua ecosystem has been limited. Not null, but still limited.

Lua itself manages to stay agnostic from all that, by staying itself dependency-free and pretty minimal. By keeping the core Teal compiler dependency-free, my intention is to dodge that same bullet. So the scope question for `tl.tl` is pretty much settled.

As for `tl`, I think that is the core question that needs to be decided: what's the scope?

1. "Compiler" - takes input files, produces output files
2. "Build system" - understands a 'project' 
3. "Package manager" - understands dependencies, deployment, packaging

In the initial PR for `build.tl` I pushed back against having that file auto-running on things like `tl check foo.tl`. What established my rationale for that was that `tl check`, `tl gen`, `tl run` are "low-level commands", you give it a file, it does the thing on the file. In that sense I see those more like "compiler" commands (well, `tl run` falling into "interpreter").

`tl build`, as its name implies is a "build system" command. It could have easily been made into a separate `tlbuild` command. The proposal of `tl init` (#363) falls along the same lines, providing a pretty opinionated view of how a Teal 'project' should look like (and I'm not sure how suitable that view is for different Lua environments, but I'm also not sure if that's an important concern).

When @lenscas talks about running luapak and the like, we start to fall into the packaging and deployment area, and honestly, the more one talks about in terms of projects, the sooner the issue of dependencies will come along and the need to integrate package management. It will happen sooner or later!

Right now, the more I think about it, the more I think `tl` should be a low-level compiler frontend (more `rustc` than `cargo`) and stick to category 1. And we need a new tool that does category 2 and possibly — or rather, eventually inevitably :) — category 3. Maybe that is https://github.com/euclidianAce/teal-cli/ or a descendant of it!

For the end-user experience, what's important is that those tools are available in a simple manner (e.g. both are installed with the officially-advertised way of installing Teal — a package with precompiled binaries sounds attractive) and we make sure in the documentation which one is the intended interface for them to use according to their use-case. For example, for an embedded use of Teal, dropping `tl.lua` into their environment and calling `tl.loader()` from Lua should work. For a greenfield Teal application, using a single full-featured CLI to create a project, manage dependencies, build and deploy would be a great experience.

Should we flip the names and make `tlc` be the low-level compiler that lives in this repo (and possibly rename the repo to github.com/teal-language/teal, to yet again take a page from Rust!) and have `tl` be the nice colorful CLI that has its own repo? :)

lenscas:
I'm down for splitting `tl` into a more low level compiler style of program and one that has a better idea of what the project looks like and how to transform the output to the desired result. Right now `tl` is already 1174 lines long and this will probably only get bigger, making it more focused would combat that. 

So, I think this PR can be closed then and a github discussion should be opened about what people want from a more higher level CLI and on whether we take the current `build` command from `tl` as a base, use https://github.com/euclidianAce/teal-cli/ or go with something else.

euclidianAce:
Both of you are making some good insights, I think I'll make a short writeup for a discussion about the current status of `tl` and `teal-cli`, where I think they're going, and make some suggestions about features to help get some more users feedback about what they expect from that sort of tool

hishamhm:
Closing this one since it's superseded by Cyan!

PR #359, opened 2021-01-26
stdlib: io.read doesn't take a file argument

euclidianAce:


hishamhm:
@euclidianAce good catch!!

PR #360, opened 2021-01-26
Build: look in parent directories for tlconfig.lua

euclidianAce:
Attempting to use the official cli for a project, forgot I hadn't ported this feature :laughing:

### DIFF
--- a/tl
+++ b/tl
@@ -43,6 +43,35 @@ local function find_in_sequence(seq, value)
    return false
 end
 
+local function str_split(str, delimiter)
+   local idx = 0
+   return function()
+      if not idx then return end
+      idx = idx + 1
+      local prev_idx = idx
+      local s_idx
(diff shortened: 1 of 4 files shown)


tl
```diff
@@ -273,6 +245,69 @@ if not cmd then
...
+      -- Merge tlconfig with the default config
+      if user_config then
+         for k, v in pairs(user_config) do
+            config[k] = v
+         end
+      end
+   end
+
+   if not config.build_file then
+      if lfs.attributes("./build.tl", "mode") == "file" then
```
hishamhm, line 287: the lfs.chdir below is happening after this check, right? It seems like this check/adjustment of the `build_file` value should be moved below it (or the lfs.chdir should be moved in here).

euclidianAce: Yep, good catch

PR #362, opened 2021-01-28
Lazy rebuilds

euclidianAce:
Implements lazy rebuilds for `tl build`
 - Adds the `-u --update-all` flag for `build` to force a complete recompilation
 - Heavily refactors `tl` to (hopefully) make it more organized
 - `tl.parse_program` now returns (as a 3rd ret val) a list of all nodes that are calls to `require` with literal string arguments so that the dependency graph can be made by only parsing each file, rather than fully type checking each file to get the dependencies. I wasn't sure if this was the best way to do it, but it seemed like the path of least resistance. (cc @hishamhm any suggestions for this? imo its either this or having `parse_program` return something like a `Result`)

Internal notes:
 - I've gotten rid of the `project` object that I originally implemented for the build command as it would eagarly scan and build, which is the opposite of what we usually want. Plus it was getting a little bloated for my tastes. It has been replaced by some more filesystem utilities. The include and exclude pattern matching facilities are now more object oriented with a type looking something like this:
```
record Matcher
   {string}
   add_pattern: function(Matcher, string)
   match: function(Matcher, string): boolean
end
```
- To replace the `traverse` function that would build up a tree of the project, there is a new `scan_dir(string, Matcher, Matcher): function(): string` iterator function, which takes a dirname, scans it and only yields the matching file names.
This is used to build up a directed acyclic dependency graph (DAG), that is used to mark files for updating, then they are simply iterated over and compiled.

 - Sadly since the build script relied decently on the `project` structure, I've had to reimplement it and take away a small amount of its functionality, but I think it mostly still functions as intended. (cc @lenscas feel free to let me know if I've ruined your workflow!) So the build script will now only run if there are files to be compiled.

 - And finally, I've attempted to add some more convenient type checking functions for checking the config and build files. The function mostly on duck typing, but its better than nothing.


euclidianAce:
Oh! and I exposed `require_module`and made it a tad more `nil`-safe since I was caching the results from `parse_program`, so I needed a way to preload modules into an environment, but I didn't want to use any of the `process` functions since I already had the ASTs generated.

euclidianAce:
One thing that I should mention is that this puts the official cli at feature parity with [my cli](https://github.com/euclidianAce/teal-cli) with the exception of fun ANSI colors and progress bars

lenscas:
> Sadly since the build script relied decently on the project structure, I've had to reimplement it and take away a small amount of its functionality, but I think it mostly still functions as intended. (cc @lenscas feel free to let me know if I've ruined your workflow!) So the build script will now only run if there are files to be compiled.

I did not yet look at the changes, but it only running if files changed is a good thing. The less often it has to run the better.

euclidianAce:
Actually after a second look, I forgot to actually implement that, adding that now!

tl.tl
```diff
@@ -1837,6 +1853,9 @@ do
       i, lhs = P(ps, i)
       i, lhs = E(ps, i, lhs, 0)
       if lhs then
+         if node_is_require_call(lhs) then
```
hishamhm, line 1856: this test can be moved inside the handler of @funcall, so that it doesn't hit for every expression

tl.tl
```diff
@@ -1837,6 +1853,9 @@ do
...
+            table.insert(ps.require_calls, lhs)
```
hishamhm, line 1857: instead of extending Node, `require_calls` could be a `{string}` and just store the contents of the literal (`lhs.e2[1].tk`)

tl.tl
```diff
@@ -4022,11 +4041,20 @@ local function fill_field_order(t: Type)
...
 
-local function require_module(module_name: string, lax: boolean, env: Env, result: Result): Type, boolean
+function tl.require_module(module_name: string, lax: boolean, env: Env, result: Result): Type, boolean
+   env = env or tl.init_env(lax)
    local modules = env.modules
    local loaded = env.loaded
 
    if modules[module_name] then
+      if result and not result.dependencies[module_name] then
+         -- TODO: should we store a map of module names to file names in Env to get rid of this redundant search?
```
hishamhm, line 4051: yes, I think so. We already have `loaded`, which is a map from filenames to Results, and `modules` which is a map from module names to types. A map of module names to file names would be essentially a join table between these two. 

However....

Don't we already have `filename` in the Type? I don't think all generated Type objects currently have that set, but we could ensure that everything returned by `require()` does. In that case, you could have your mapping from module name to filename by doing `env.modules[module_name].filename`.

hishamhm (commented):
I only reviewed `tl.tl` since these are the compiler changes needed for the feature (as the build features themselves could end up elsewhere, as per the discussion going on [here](https://github.com/teal-language/tl/pull/358#issuecomment-769388255)) — overall, the changes look fine to me, happy to add whatever improvements are needed to the compiler to make the build functionalities work!

euclidianAce:
I want to say this pr is done, or at least implements what I had planned, I think the next big thing is to figure out the scope of it and where it's going to live. One thing I'd definitely like for the language server is just to be able to expose the `get_config()` function so the server can be aware of the entire project and its configuration.

hishamhm (commented):
> I want to say this pr is done, or at least implements what I had planned, I think the next big thing is to figure out the scope of it and where it's going to live. 

Looks good, but it's really looking more like a build system now, I think it really makes sense to split it from the low-level compiler frontend.

> One thing I'd definitely like for the language server is just to be able to expose the get_config() function so the server can be aware of the entire project and its configuration.

...which would allow it to be written in Teal and expose any necessary submodules and APIs, like your tl-cli project. I should have gone to sleep hours ago, but let's sort this out as discussed in https://github.com/teal-language/tl/pull/358 and https://github.com/teal-language/tl/discussions/371.

tl.tl
```diff
@@ -4022,18 +4046,22 @@ local function fill_field_order(t: Type)
...
+         result.dependencies[module_name] = env.modules[module_name].filename
+      end
       return modules[module_name], true
    end
    modules[module_name] = UNKNOWN
 
    local found, fd = tl.search_module(module_name, true)
    if found and (lax or found:match("tl$") as boolean) then
-      fd:close()
+      (fd as FILE):close()
```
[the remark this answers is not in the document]
euclidianAce: ah, yeah, I was running into an issue that I think is either the same as #365 or related to it, getting the `argument 0: FILE is not a FILE` error when loading `tl.tl` as a module while working on the language server

euclidianAce:
Boy a lot of stuff got added when I wasn't looking :laughing:. Attempted to rebase this over the current master and everything seems to be in working order. A commit or two might be out of place though

hishamhm:
I'm pushing a new release today, so I'm cherry-picking the `tl.tl` changes from this PR to get them merged!

tl.tl
```diff
@@ -4897,11 +4935,12 @@ tl.type_check = function(ast: Node, opts: TypeCheckOptions): {Error}, {Error}, T
...
-      syntax_errors = {},
-      type_errors = {},
-      unknowns = {},
-      warnings = {},
+   local result = {
+      syntax_errors = opts.result and opts.result.syntax_errors or {},
+      type_errors = opts.result and opts.result.type_errors or {},
+      unknowns = opts.result and opts.result.unknowns or {},
+      warnings = opts.result and opts.result.warnings or {},
+      dependencies = opts.result and opts.result.dependencies or {},
```
hishamhm, line 4943: I have passed around these lists from `result` to `result` in order to combine errors from multiple files into a single list. If you want each file's `result` to have only its list of direct dependencies, I think each `result` needs a fresh `dependencies` table.

hishamhm:
> Oh! and I exposed require_moduleand made it a tad more nil-safe since I was caching the results from parse_program, so I needed a way to preload modules into an environment, but I didn't want to use any of the process functions since I already had the ASTs generated.

I'm okay with merging this as it doesn't affect the regular compiler flow, but... should we really make `env` and `result` optional? If the caller doesn't pass those variables (`env` especially), the version in this PR will `init_env` every time, nothing will be cached, and it _will_ trigger `tl.process`...

euclidianAce:
Closing this as the functionality is being put in a new tool

PR #367, opened 2021-01-29
perform flow inference based on ==

hishamhm:
Makes code such as this valid:

```
local x: number | string

if x == 9 then
   print(x + 2)
end
```

A more interesting use is propagating string constants into enums:

```
local enum Direction
   "north"
   "south"
   "east"
   "west"
end

local function f(d: Direction)
   print(d)
end

local s: string
if s == "north" or s == "south" then
   f(s)
end
```

Note that _assignments_ do not (yet) propagate, since that would require a proper constant propagation algorithm. This is more of a clever hack reusing the conservative flow analysis I had implemented for `is`.

### DIFF
--- a/spec/operator/eq_spec.lua
+++ b/spec/operator/eq_spec.lua
@@ -79,6 +79,62 @@ describe("flow analysis with ==", function()
          { msg = "cannot index something that is not a record: number | string" }
       }))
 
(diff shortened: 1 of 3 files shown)


hishamhm:
I'm going to go ahead and merge this and see what happens. If any feedback comes up, we'll adjust the inference rules accordingly!

PR #369, opened 2021-01-31
tl types: new CLI command (and tl API) for dumping type information

hishamhm:
This is an MVP for dumping type information for consumption by third-party tools.

These were my high-level goals:

* minimal tl API
* not exposing the raw AST
* simple output, designed for quick lookup
* JSON output in the CLI

Right now the main thing is a map mapping filename:y:x positions to types.

One next improvement to add would be mapping variables to their definition locations (right now they map to their types only).

By visual inspection alone, I have already noticed that some of the information returned is a bit wonky and can be improved (some nested AST nodes report the same x, y positions and I just picked the first one; so we probably should tweak the AST a bit to ensure the most relevant information is returned). But I think this is a good start, and should be a good kickstart for getting feedback on what data is missing and what can work better!

(Visualizing the raw data this way has already prompted some other improvements in the compiler, even, see the other commits in this PR!)

### DIFF
--- a/spec/call/syntax_errors_spec.lua
+++ b/spec/call/syntax_errors_spec.lua
@@ -5,7 +5,6 @@ describe("call", function()
       print("hello", "world",)
    ]], {
       { msg = "unexpected ')'" },
-      { msg = "syntax error, expected ')'" },
    }))
 
    it("cannot call a string", util.check_syntax_error([[
--- a/spec/cli/types_spec.lua
+++ b/spec/cli/types_spec.lua
@@ -1,0 +1,155 @@
+local assert = require("luassert")
+local util = require("spec.util")
+
+describe("tl types works like check", function()
+   describe("on .tl files", function()
+      it("works on empty files", function()
+         local name = util.write_tmp_file(finally, [[]])
+         local pd = io.popen(util.tl_cmd("types", name), "r")
+         local output = pd:read("*a")
+         util.assert_popen_close(true, "exit", 0, pd:close())
+         -- TODO check json output
+      end)
+
+      it("reports 0 errors and code 0 on success", function()
+         local name = util.write_tmp_file(finally, [[
+            local function add(a: number, b: number): number
+               return a + b
+            end
+
+            print(add(10, 20))
+         ]])
+         local pd = io.popen(util.tl_cmd("types", name), "r")
+         local output = pd:read("*a")
+         util.assert_popen_close(true, "exit", 0, pd:close())
+         -- TODO check json output
+      end)
+
+      it("reports number of errors in stderr and code 1 on type errors", function()
+         local name = util.write_tmp_file(finally, [[
+            local function add(a: number, b: number): number
+               return a + b
+            end
+
+            print(add("string", 20))
+            print(add(10, true))
+         ]])
+         local pd = io.popen(util.tl_cmd("types", name) .. "2>&1 1>/dev/null", "r")
+         local output = pd:read("*a")
+         util.assert_popen_close(nil, "exit", 1, pd:close())
+         assert.match("2 errors:", output, 1, true)
+         -- TODO check json output
+      end)
+
+      it("reports errors in multiple files", function()
+         local name1 = util.write_tmp_file(finally, [[
+            local function add(a: number, b: number): number
+               return a + b
+            end
+
+            print(add("string", 20))
+            print(add(10, true))
+         ]])
+         local name2 = util.write_tmp_file(finally, [[
+            local function add(a: number, b: number): number
+               return a + b
+            end
+
+            print(add("string", 20))
+            print(add(10, true))
+         ]])
+         local pd = io.popen(util.tl_cmd("types", name1, name2) .. " 2>&1 1>/dev/null", "r")
+         local output = pd:read("*a")
+         util.assert_popen_close(nil, "exit", 1, pd:close())
+         assert.match(name1 .. ":", output, 1, true)
+         assert.match(name2 .. ":", output, 1, true)
+         -- TODO check json output
+      end)
+
+      it("reports number of errors in stderr and code 1 on syntax errors", function()
+         local name = util.write_tmp_file(finally, [[
+            print(add("string", 20))))))
+         ]])
+         local pd = io.popen(util.tl_cmd("types", name) .. "2>&1 1>/dev/null", "r")
+         local output = pd:read("*a")
+         util.assert_popen_close(nil, "exit", 1, pd:close())
+         assert.match("1 syntax error:", output, 1, true)
+         -- TODO check json output
+      end)
+
+      it("reports use of unknowns as errors in stderr and returns code 1", function()
+         local name = util.write_tmp_file(finally, [[
+            local function unk(x, y): number, number
+               return a + b
+            end
+         ]])
+         local pd = io.popen(util.tl_cmd("types", name) .. "2>&1 1>/dev/null", "r")
+         local output = pd:read("*a")
+         util.assert_popen_close(nil, "exit", 1, pd:close())
+         assert.match("2 errors:", output, 1, true)
+         assert.match("unknown variable: a", output, 1, true)
+         assert.match("unknown variable: b", output, 1, true)
+         -- TODO check json output
+      end)
+   end)
+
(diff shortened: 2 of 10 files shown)


pdesaulniers:
Very cool!

Right now, I'm trying to use the `types` command to implement autocompletion in vscode-teal. It seems like the command crashes on partial input sometimes:

```lua
local record Point
    x: number
    y: number
end

local
```

```
tl -q types test.tl 
1 syntax error:
test.tl:8:6: expected a local variable definition
/usr/bin/lua5.4: tl:857: attempt to index a nil value (local 'result')
stack traceback:
        tl:857: in main chunk
        [C]: in ?
```

hishamhm:
> Right now, it seems like the command crashes on partial input sometimes:

Ah, dealing with partial input might prove be super tricky, because currently the compiler doesn't run the type checker in the presence of syntax errors... the language server may get away with caching the type information from the last time it succeeded type checking.

Partial inputs is often one of the reasons why LSP integrations end up using their own parser rather than using the compiler's. I'm not sure how much they also tend to re-do the compiler's work with regard to symbol tables and scopes in order to provide auto-completion, or how much the compiler is expected to provide this information.

pdesaulniers:
Regarding the `types` command, I've noticed one case where the mapping doesn't seem right.

I'm trying to go to the definition of the `Point` type at `test.tl:3:15`

test.tl
```lua
local Point = require("point")

local x = {x: Point = {x = 1, y = 2}, y = 2}
```

point.tl
```lua
local record Point
    x: number
    y: number
end

return Point
```

The JSON data looks like this:
```json
{
    "types": {
        "1": {
            "str": "()"
        },
        "2": {
            "str": "function(string)"
        },
        "3": {
            "x": 23,
            "str": "string",
            "y": 1
        },
        "4": {
            "str": "(string)"
        },
        "5": {
            "str": "(type record)"
        },
        "6": {
            "x": 28,
            "str": "number",
            "y": 3
        },
        "7": {
            "str": "record",
            "file": "test.tl",
            "fields": {
                "x": 6,
                "y": 6
            },
            "x": 23,
            "y": 3
        },
        "8": {
            "str": "(record)"
        }
    },
    "by_pos": {
        "test.tl": {
            "1": {
                "15": 2,
                "13": 5,
                "22": 4,
                "23": 3
            },
            "3": {
                "35": 6,
                "11": 7,
                "39": 3,
                "24": 3,
                "9": 8,
                "7": 1,
                "43": 6,
                "12": 3,
                "28": 6,
                "23": 7,
                "31": 3
            }
        }
    }
}
```

It seems like there is no data for `test.tl:3:15`. 

However, if I tweak the example like this:

```lua
local Point = require("point")

local x: Point = {x = 1, y = 2}
```

then, I can access the Point record declaration via `test.tl:3:10` (by going to type "9", then "10" via the "ref" field):

```json
{
    "types": {
        "1": {
            "str": "()"
        },
        "2": {
            "str": "function(string)"
        },
        "3": {
            "str": "string",
            "y": 1,
            "x": 23
        },
        "4": {
            "str": "(string)"
        },
        "5": {
            "str": "(type record)"
        },
        "6": {
            "str": "number",
            "y": 3,
            "x": 23
        },
        "7": {
            "file": "test.tl",
            "y": 3,
            "x": 18,
            "str": "record",
            "fields": {
                "y": 6,
                "x": 6
            }
        },
        "8": {
            "str": "(record)"
        },
        "9": {
            "file": "test.tl",
            "y": 3,
            "x": 10,
            "str": "Point",
            "ref": 10
        },
        "10": {
            "file": "./point.tl",
            "y": 1,
            "x": 1,
            "str": "type record",
            "fields": {
                "y": 6,
                "x": 6
            }
        },
        "11": {
            "str": "(Point)",
            "file": "test.tl",
            "y": 3,
            "x": 8
        }
    },
    "by_pos": {
        "test.tl": {
            "1": {
                "15": 2,
                "13": 5,
                "22": 4,
                "23": 3
            },
            "3": {
                "16": 8,
                "18": 7,
                "19": 3,
                "23": 6,
                "8": 11,
                "10": 9,
                "7": 1,
                "30": 6,
                "26": 3
            }
        }
    }
}
```

hishamhm:
@pdesaulniers The discrepancy you noted was indeed a bug in the compiler! Commit 1848cd0 should fix that one specifically.

I pushed some other tweaks to improve the reported data and locations.

hishamhm:
All right, I made a bunch of tweaks to tl-types branch, and also made one BIG experimental change: now the type checker keeps going **even in the event of syntax errors**. I try to skip them over, resync, and still build a valid AST (which now contains a new `"error_node"` which produces `<invalid type>`, to fill in the blanks). This is barely tested and there might be tons of bugs because all of the typechecking step code was written assuming the AST was 100% valid, and now there may be missing nodes that need to be fixed in the parser code.

I played a bit with my personal Dit integration and I got "hover to show type" and "go to function definition" (mostly) working. (it's "mostly" working because when a function is forward declared with a local, it goes to the forward declaration, not the assigned body, which would be more useful — this is because this was not really implemented as "go to variable definition", but rather "go to type definition"...)

pdesaulniers:
I just noticed that `tl types` crashes on empty files:

```
tl.lua:8246: attempt to index a nil value (field '?')
stack traceback:
        tl.lua:8246: in function 'tl.get_types'
        tl:897: in main chunk
        [C]: in ?
```

hishamhm:
@pdesaulniers good catch! I guess it's time to start writing actual tests for the `tl types` functionality!...

pdesaulniers:
This line of code crashes `tl types`:

```lua
local function do
```

```
/usr/bin/lua5.4: /home/patrick/.luarocks/share/lua/5.4/tl.lua:6351: attempt to index a nil value (field 'name')
stack traceback:
        /home/patrick/.luarocks/share/lua/5.4/tl.lua:6351: in upvalue 'add_function_definition_for_recursion'
        /home/patrick/.luarocks/share/lua/5.4/tl.lua:7558: in local 'bs'
        /home/patrick/.luarocks/share/lua/5.4/tl.lua:2843: in upvalue 'extra_callback'
        /home/patrick/.luarocks/share/lua/5.4/tl.lua:2936: in upvalue 'recurse_node'
        /home/patrick/.luarocks/share/lua/5.4/tl.lua:2863: in upvalue 'recurse_node'
        /home/patrick/.luarocks/share/lua/5.4/tl.lua:8089: in function 'tl.type_check'
        /home/patrick/.luarocks/share/lua/5.4/tl.lua:8400: in function 'tl.process_string'
        /home/patrick/.luarocks/share/lua/5.4/tl.lua:8344: in function 'tl.process'
        ...patrick/.luarocks/lib/luarocks/rocks-5.4/tl/dev-1/bin/tl:663: in local 'type_check_file'
        ...patrick/.luarocks/lib/luarocks/rocks-5.4/tl/dev-1/bin/tl:895: in main chunk
        [C]: in ?
```

pdesaulniers:
@hishamhm It seems like `tl types` is returning some weird data for `ParseItem` in the latest `tl.tl` from master (but not `tl.tl` from this branch, strangely!)

If I go to the definition of that type and create a local variable `abcdefg` of type `ParseItem` right below:

```lua
local type ParseItem = function<T>(ParseState, number, number): number, T, number

local abcdefg: ParseItem = nil
```
...then, `tl types` reports that the type of `ParseItem` next to `abcdefg` is the `string` table? :)

```
        "329": {
            "fields": {
                "byte": 332,
                "gmatch": 333,
                "lower": 334,
                "pack": 335,
                "find": 336,
                "unpack": 337,
                "sub": 347,
                "packsize": 339,
                "match": 340,
                "dump": 341,
                "rep": 342,
                "char": 343,
                "format": 344,
                "gsub": 345,
                "reverse": 348,
                "len": 338,
                "upper": 346
            },
            "str": "record"
        },
```

hishamhm:
@pdesaulniers `tl-types` branch now includes globals in the report! The left-hand-side of declarations should have the type included as well. The crash above should be fixed. Also, I think the issue you saw was caused by unstable traversal producing different numbers for the same input; this should be improved now (there might be glitches but I believe it is stable), but note that whenever the _input_ changes, the numbers are not guaranteed to be stable.

PR #384, opened 2021-02-08
Close top level scope at the end of the file

pdesaulniers:
This fixes the output of `tl types -p y:x` when `y:x` is located after the last token in the file.

Some of the tests assumed that the errors would point to the last token in the file instead of the EOF, so I trimmed the trailing whitespace in a few places in `spec/util.lua`.

There are now 2 tests that fail because of the lack of trailing whitespace (these errors can be reproduced on master):

1) This line returns `variable 'x' has no type or initial value` instead of `expected an expression`

```lua
local x =
```

2) This line returns `syntax error, expected '='` instead of `expected a type`

```lua
local type argh =
```

### DIFF
--- a/spec/lexer/eof_spec.lua
+++ b/spec/lexer/eof_spec.lua
@@ -11,6 +11,6 @@ end
 describe("lexer", function()
    it("equals at the end of a line", function()
       local tokens = tl.lex("local type argh =")
-      assert.same({"local", "type", "argh", "="}, map(function(x) return x.tk end, tokens))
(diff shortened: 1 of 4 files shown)


hishamhm:
> There are now 2 tests that fail because of the lack of trailing whitespace (these errors can be reproduced on master):

I can't reproduce those on master, I got "expected an expression" and "expected a type" respectively.

hishamhm (approved):

PR #389, opened 2021-02-10
fixed math.randomseed() types; both input and returned

JLPLabs:
math.randomseed accepts two values and returns two values. 
tl.tl was updated to reflect this. (the original version only had 1 input and 0 return values)


hishamhm:
@JLPLabs it's now merged and will feature in the next release, thank you for making the PR!

PR #392, opened 2021-02-13
Track require dependencies in parser (Cherry picks from #362)

hishamhm:
This PR contains the parser changes that track `require` calls as it parses, plus other minar changes outside of `tl` and `spec/cli`from #362.

@euclidianAce I have the commit that exposes `require_module` ready as well, but I chose not to add it because I wanted to discuss that a bit further: is it really necessary? From what I could follow in #362 it seems it was exposed in order to do "preload modules" outside of `tl.process`. I'm thinking it might make more sense to move the "preload modules" functionality into `tl.init_env` instead, what do you think?

### DIFF
--- a/spec/api/parse_program_spec.lua
+++ b/spec/api/parse_program_spec.lua
@@ -1,0 +1,17 @@
+local tl = require("tl")
+
+describe("tl.parse_program", function()
+   it("returns a list of all require arguments", function()
+      local tl_code = [[
+         require("foo")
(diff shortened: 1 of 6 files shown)


euclidianAce:
> I'm thinking it might make more sense to move the "preload modules" functionality into tl.init_env instead, what do you think?

Yes! that would make it much easier for things not using the `process` functions

PR #395, opened 2021-02-14
Fix Teal Playground link

bew:
Otherwise the link generated by Github is `https://github.com/teal-language/tl/blob/master/teal-playground.netlify.app` which obviously doesn't exist


hishamhm:
Oops! Thank you!!

PR #397, opened 2021-02-17
tl types: set "file" field for function types

pdesaulniers:
This PR adds a `file` field to function types in `tl types`'s data.

Appears to fix https://github.com/teal-language/vscode-teal/issues/32.


hishamhm:
@pdesaulniers merged! I just renamed the commit to explain the change a bit because at first it wasn't obvious to me how the change in the filename field was adding the file field in `tl types` but I read through the code and figured it out :)

PR #398, opened 2021-02-18
Make function types look less compact

pdesaulniers:
This PR tweaks the output of `show_type_base` for function types. It makes the type strings look more evenly spaced.

Before:
```lua
function(number,function(number,{string : string})):string,number
```
After:
```lua
function(number, function(number, {string : string})): string, number
```


hishamhm:
@pdesaulniers merged! :+1:

PR #400, opened 2021-02-19
Changes in the resolution of nominals

### LINKED ISSUE #269: Using the type declaration syntax prevents nominals from being properly resolved
minimal example:
```
local type A = number
local type B = A

local function foo(bar: A)
end

local var: B
foo(var)
```
`tl check`:
```
1 error:
blah.tl:9:5: argument 1: B is not a A
```
except that `B` is exactly `A`, so i would imagine that this should work.
Additionally, i believe this is what prevents us from using `local type ... = ...` here
https://github.com/teal-language/tl/blob/e769dc803e8eb06f4e1f16489d01f92a02dc2c19/tl.tl#L48-L50
Which would help clean up generated code, which shouldn't have any traces of non-record definitions

hishamhm:
* Avoid producing nominal types that map to other nominal types; `resolve_tuple_and_nominal` (formerly `resolve_unary`) always results in a non-nominal.
* Nominals without type arguments declared with `local type` or `global type` now work as aliases. That means that if you do `local type Foo2 = Foo` then `Foo2` is the same type as `Foo`.
* For now at least, type constructors with arguments resolve to new distinct types. That means that if you do `local type FooNumber = Foo<number>` then `FooNumber` is a new distinct type from `Foo<number>` and not an alias.

Fixes #269.

### DIFF
--- a/spec/declaration/global_spec.lua
+++ b/spec/declaration/global_spec.lua
@@ -199,4 +199,53 @@ describe("global", function()
       end)
    end)
 
+   describe("with types", function()
+      it("nominal types can take type arguments", util.check [[
+         global record Foo<R>
+            item: R
+         end
+
+         global type Foo2 = Foo
+         global type Bla = Foo<number>
+
+         global x: Bla = { item = 123 }
+         global y: Foo2<number> = { item = 123 }
+      ]])
+
+      it("nested types can be resolved as aliases", util.check [[
+         global record Foo<R>
+            enum LocalEnum
+               "loc"
+            end
+
+            record Nested
+               x: {LocalEnum}
+               y: R
+            end
+
+            item: R
+         end
+
+         global type Nested = Foo.Nested
+      ]])
+
+      it("types declared as nominal types are aliases", util.check [[
+         global record Foo<R>
+            item: R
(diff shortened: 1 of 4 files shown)


hishamhm:
Absorbed by #402

euclidianAce:
I think I've found a case where the behavior has changed and I'm not sure if its correct or not.
It has to do with nested records being put in tables:
```lua
-- foo.tl
local record Foo
   record Bar
   end
end
local function func(_f: Foo.Bar) end

local tab = { Foo = Foo }

local x: tab.Foo.Bar
func(x)
```
```
$ tl check foo.tl 
1 error:
foo.tl:11:6: argument 1: tab.Foo.Bar is not a Foo.Bar
```

It works for non-nested records, e.g.
```lua
local function func(_f: Foo) end
local x: tab.Foo
func(x)
```
type checks just fine. Is this behavior expected? Is storing types in tables like above a valid use case (for things like modules) or should it be discouraged/disallowed?

hishamhm:
@euclidianAce This as an actual bug! I'm looking into it!

hishamhm:
Fixed in master!

PR #403, opened 2021-02-20
fix generation of symbols list and minimize it

hishamhm:


hishamhm:
@pdesaulniers This should fix the results of `tl types`!

pdesaulniers:
Yes, I think it does! The symbol list appears to be sorted correctly now :+1:

PR #404, opened 2021-02-24
[wip] tl.tl: Include a tk: string field in tl.Error + expose Token

euclidianAce:
Most of the token information was already present in tl.Error, now just
include the string as well for tooling like cyan, so they don't have to
search for the token to get the length/content of it.

Also expose Token and TokenKind as tl.Token and tl.TokenKind since they
are relatively simple types.

### DIFF
--- a/spec/declaration/local_spec.lua
+++ b/spec/declaration/local_spec.lua
@@ -118,8 +118,8 @@ describe("local", function()
 
          assert.same({}, result.syntax_errors)
          assert.same({
-            { y = 3, x = 42, filename = "main.tl", msg = "in local declaration: var: unknown field dato" },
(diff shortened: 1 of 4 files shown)


euclidianAce:
after working with this for a little bit, it seems that the ast (being abstract and all) sometimes doesn't carry the "correct" token with it in certain nodes and produces weird results. I still think this is worth having but I'll have to work a little more on it

hishamhm:
Maintaining this contextual information is often tricky. For your use case, instead of exposing more internals from the compiler, I think it would be better to just add and expose a single function `function tl.get_token(tokens: {Token}, y: number, x: number): string`

euclidianAce:
Yeah, both cyan and the language server implement a `get_token`. I was hoping it might be a bit easier to maintain context but after trying it that does seem like the better solution

Issue #195, opened 2020-07-24
How to use types that aren't explicitly exposed in a module


euclidianAce:
I'm working on some tooling for neovim in the absence of a language server and am using `tl.tl` as an api and am having trouble defining my own types using some of the local types like `Env`.

I have this type:
```
local TypeChecker = record
   --nvim
   buffer: number
   namespaceID: number

   --tl
   config: table
   env: table -- tl.Env isn't a thing
end
function TypeChecker:init()
   -- [[ nvim stuffs ]]
   -- load config from current dir
   self.env = tl.init_env(false, self.config.skip_compat53)
end
```
I'd like to attach the env to the object for ease of using `tl.process_string`, which requires the `Env` type as an argument, which contains a multitude of other local types so I can't just use an `as {string:any}` since `tl check` complains.

Is this just something that should be the responsibility of the module to make sure and expose the type? Or should local types not be allowed to be exposed like this? I can work around this by calling `tl.init_env` each time I call `tl.process_string` but that doesn't seem right.

I know you're working on breaking `tl.tl` into modules and I think this is a good thing to think about while doing so :)

PR #405, opened 2021-02-24
tl.tl: add tl.get_token_at

euclidianAce:
Just a binary search on a (sorted) list of tokens to find the one at the given location, returns the Token at that location so the Token type is also exposed as `tl.Token`

### DIFF
--- a/tl.tl
+++ b/tl.tl
@@ -826,29 +826,46 @@ do
    end
 end
 
-function tl.get_token_at(tks: {Token}, y: number, x: number): Token
-   local len <const> = #tks
-   local max_steps <const> = math.ceil(math.log(len, 2)) + 1
-   local guess = math.floor(len / 2)
-   local factor = math.floor(len / 4)
-   for _ = 1, max_steps do
-      local tk <const> = tks[guess]
-      local tk_x <const>, tk_y <const> = tk.x, tk.y
-
-      local sign <const> = tk_y > y  and -1
-         or tk_y < y and 1
-         or tk_x <= x and x < tk_x + #tk.tk and 0
-         or tk_x > x and -1
-         or 1
-
-      if sign == 0 then
-         return tk
+local function binary_search<T, U>(list: {T}, item: U, cmp: function(T, U): boolean): number, T
+   local len <const> = #list
+   local mid: number
+   local s, e = 1, len
+   while s <= e do
+      mid = math.floor((s + e) / 2)
+      local val <const> = list[mid]
+      local res <const> = cmp(val, item)
+      if res then
+         if mid == len then
+            return mid, val
+         else
+            if not cmp(list[mid + 1], item) then
+               return mid, val
+            end
+         end
+         s = mid + 1
       else
-         guess = guess + sign * factor
+         e = mid - 1
       end
+   end
+end
 
-      guess = math.min(math.max(guess, 1), len)
-      factor = math.max(math.floor(factor / 2), 1)
+function tl.get_token_at(tks: {Token}, y: number, x: number): Token
+   local _, found <const> = binary_search(
+      tks, nil,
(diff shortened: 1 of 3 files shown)


euclidianAce:
I just noticed the floor division operators didn't get converted (presumably because im running 5.3) should the config target 5.1 for max compatability?

hishamhm:
> I just noticed the floor division operators didn't get converted (presumably because im running 5.3) should tl target 5.1 for max compatability?

To avoid bootstrapping annoyances, just use `math.floor`, otherwise we'll have to be extra careful to always target 5.1. Teal tries to paper over Lua version differences to make it less of an issue for _Teal users_, but for _Teal compiler developers_ themselves, we need to take the hit and remain careful and stick to the common subset to make `tl.lua` universal (just using `--gen-target=5.1` is not enough because it may generate things that are not right for 5.3+)

hishamhm:
Also, math.log does not support a base argument in Lua 5.1. I guess you could convert the bases over, but at that point would it be simpler to just use a while loop? Actually, now that I think of it, there's already another implementation of binary search in `tl.tl` (`find`, inside `symbols_in_scope`). Ideally, that one could be factored out as a `local function `binary_search<T>(list: {T}, item: T, cmp: function(T, T): boolean): T` instead of adding a second implementation?

tl.tl
```diff
@@ -458,6 +460,9 @@ do
 
       local function end_token_prev(kind: TokenKind)
          local tk = input:sub(ti, i - 1)
+         if kind == "identifier" and keywords[tk] then
```
hishamhm, line 463: This check shouldn't be added here. The reason why I split `end_token` into multiple tokens was precisely to specialize their behavior and avoid this extra test in all non-`identifier` states.

I suspect the reason you added this here is because of the test for the very last token that happens near the end of `lex` which is assuming `end_token_prev`. Instead of adding a test here that would add a cost for every token, that final test should instead check which of `end_token_prev` or `end_token_identifier` should be called, based on the current `state`.

hishamhm:
@euclidianAce Thank you for applying the changes! One last change, but one I can make it myself when merging since I'm at the computer now (some of the reviews above were made by phone :grin: ), is to not expose the `Token` and `TokenKind` types (in particular, I don't feel `TokenKind` is "stable" yet). That's why my suggestion was to return only the `tk` string from the function. 

I assume you wanted the TokenKind in order to do syntax highlighting? That wouldn't be very useful, because soft keywords such as `record` are not included, and for full correctness these words should be highlighted based on parse context (e.g. `type` in `type(var)` vs `local type Foo`). But honestly, for highlighting error message lines in Cyan I wouldn't bother with that and just hardcode a list including all hard and soft keywords — it's not that bad to highlight `type` in bold in `type(var)`, after all, and things like `global: number` in a record attribute are unlikely, and these highlighting mishaps would certainly be forgiven by the user.

euclidianAce:
Part of the reason was to be able to annotate types in cases where inference doesn't quite cut it
```
local token_cache: {string : {tl.Token}} = {}
token_cache["foo.tl"] = tl.lex("...")
```
and more of a general problem with types being indirectly exposed through function returns (I opened #195 for the same reason). I've thought about introducing a warning for this by checking what types a module can consume and produce and what types are directly exposed then emitting a warning when a type isn't exposed but can be consumed or produced but checking it can be a little complex

I'm fine with keeping the Token type unexposed for now, but I think in general if a module can produce a type, it should be able to be annotated

hishamhm:
@euclidianAce Yes, #195 is the more general problem, and needs a proper solution (other than exposing all types). There needs to be a way to define abstract data types (expose a type without exposing its internals), so that modularity can be preserved. I've been thinking a lot about this recently.

hishamhm:
Merged manually!

PR #406, opened 2021-02-26
Integer type

hishamhm:
## Current status

As of 9c6b241 :

* `integer` is a new type
  * Integer constants without a dot (e.g. `123`) are of type `integer`
  * operations preserve the integer type according the Lua 5.4 semantics (e.g. `integer` + `integer` = `integer`, `integer` + `number` = `number`, etc.)
  * an `integer` is accepted everywhere a `number` is expected
  * variables (and type variables) **are inferred** as `integer` when initialized with integers
    * this can happen directly (`local x = 1`) or indirectly (`local x = math.min(i, j)`)
    * to infer a `number`, use `.0`: `local x = 1.0` infers a `number`
  * if a variable is of type `integer`, then an arbitrary `number` is not accepted
    * `local x: integer = 1.0` doesn't work, even though the floating-point number has a valid integer representation
* the standard library was annotated for integers
  * it only _returns_ integers, but it always accepts floats in arguments that must be integral
  * this will not change. as it matches the Lua standard library implementation: 
    * wherever you need to pass an integral value, you can pass an equivalent float with an integer representation, e.g. `table.insert(t, 1.0, "hello")` — the validity of the float is not checked at compile time, this is a Lua standard library API check

### DIFF
--- a/spec/arguments/array_spec.lua
+++ b/spec/arguments/array_spec.lua
@@ -11,8 +11,8 @@ describe("array argument", function()
          a(100)
       end
    ]], {
-      { y = 6, msg = 'expected an array: at index 2: got number, expected string' },
-      { y = 7, msg = 'argument 1: got number, expected {string}' },
+      { y = 6, msg = 'expected an array: at index 2: got integer, expected string' },
+      { y = 7, msg = 'argument 1: got integer, expected {string}' },
    }))
 
    it("constructs type of complex array correctly (#111)", util.check_type_error([[
--- a/spec/arguments/boolean_spec.lua
+++ b/spec/arguments/boolean_spec.lua
@@ -41,7 +41,7 @@ describe("boolean argument", function()
       f(function() end)
    ]], {
       { y = 12, msg = 'argument 1: got string "hello", expected boolean' },
-      { y = 13, msg = 'argument 1: got number, expected boolean' },
+      { y = 13, msg = 'argument 1: got integer, expected boolean' },
       { y = 14, msg = 'argument 1: got {}, expected boolean' },
       { y = 15, msg = 'argument 1: got R, expected boolean' },
       { y = 16, msg = 'argument 1: got function(), expected boolean' },
--- a/spec/assignment/to_array_spec.lua
+++ b/spec/assignment/to_array_spec.lua
@@ -7,8 +7,8 @@ describe("assignment to array", function()
       a = 100
       a = {"a", 100}
    ]], {
-      { y = 2, msg = "got number, expected {string}" },
-      { y = 3, msg = "expected an array: at index 2: got number, expected string" },
+      { y = 2, msg = "got integer, expected {string}" },
+      { y = 3, msg = "expected an array: at index 2: got integer, expected string" },
    }))
 
    it("resolves arity of function returns", util.check [[
--- a/spec/assignment/to_nominal_arrayrecord_spec.lua
+++ b/spec/assignment/to_nominal_arrayrecord_spec.lua
@@ -80,6 +80,6 @@ describe("assignment to nominal arrayrecord", function()
       end
       local x: Node = 123
    ]], {
-      { msg = "in local declaration: x: got number, expected Node" }
+      { msg = "in local declaration: x: got integer, expected Node" }
    }))
 end)
--- a/spec/assignment/to_nominal_record_field_spec.lua
+++ b/spec/assignment/to_nominal_record_field_spec.lua
@@ -22,7 +22,7 @@ describe("assignment to nominal record field", function()
       local t: Type = {}
       t.node = 123
    ]], {
-      { msg = "in assignment: got number, expected Node" }
+      { msg = "in assignment: got integer, expected Node" }
    }))
 
    it("fails with incorrect literal index", util.check_type_error([[
--- a/spec/assignment/to_nominal_record_spec.lua
+++ b/spec/assignment/to_nominal_record_spec.lua
@@ -63,7 +63,7 @@ describe("assignment to nominal record", function()
       end
       local x: Node = 123
    ]], {
-      { msg = "in local declaration: x: got number, expected Node" }
+      { msg = "in local declaration: x: got integer, expected Node" }
    }))
 
    it("type system is nominal: fails if different records with compatible structure", util.check_type_error([[
--- a/spec/assignment/to_self_record_field_spec.lua
+++ b/spec/assignment/to_self_record_field_spec.lua
@@ -18,6 +18,6 @@ describe("assignment to self record field", function()
          self.foo = 12
       end
    ]], {
-      { msg = "in assignment: got number, expected string" }
+      { msg = "in assignment: got integer, expected string" }
    }))
 end)
--- a/spec/call/generic_function_spec.lua
+++ b/spec/call/generic_function_spec.lua
@@ -68,9 +68,9 @@ describe("generic function", function()
          return id(cvt(x))
       end
 
-      print(use_conv(122, convert_num_str) .. "!")
+      print(use_conv(122.0, convert_num_str) .. "!")
 
-      print(use_conv("123", convert_str_num) + 123)
+      print(use_conv("123", convert_str_num) + 123.0)
    ]])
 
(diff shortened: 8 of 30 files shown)


hishamhm:
Still in Draft stage, because there are no tests for the integer type whatsoever!

hishamhm:
Some more notes:

> no currently working Teal code should stop working

This is indeed the current state of this PR, but it's not a hard guarantee (we're 0.x after all). I did change the return of things like the `ipairs` iterator, so if you try to do something funny like reassign the iterator variable with an arbitrary `number`, it might complain.

The compiler and testsuite are running without changes on this commit (c1bc529), but ymmv.

In any case I don't think this is going to stay like this:

> variables are never inferred as integer, only number

I implemented this as something open for discussion, but after a little while using this branch to start converting the compiler I'm getting convinced that this behavior is not ergonomical. Not inferring integers causes integer-based code (such as a compiler, which barely uses floats) to be full of type annotations, as each time you create a new variable passing the result of a calculation, it gets inferred to number again by default and you have to annotate.

Making things like `local x = 0` default to `integer` may make a lot of sense for integer-heavy code, but when you switch to a compiler that uses it, the integers propagate like wildfire, causing a ton of errors on existing code which has `number` declarations in various places. This is not necessarily a bad thing, but it's a major change.

The reason why I implemented this "opt-in" version first is because I generally like to implement a feature, and only after I have a commit where it works, then I switch the compiler itself to use it (so that every commit of the compiler can be built on the previous one — that's not always true, especially in the case of bugs, but I generally try to preverse the "bootstrapping history"). So mainly that's why I made it "completely non-breaking" at first, but after using it a little bit with integers, it doesn't feel like the right way to go. 

I'm working on making integers infer normally and propagate, and converting the compiler to use them. I'll let you know the results of this experiment.

hishamhm:
All right, so I successfully did the gradual conversion of the compiler to used inferred integers. The end result is pretty nice, essentially the code looks the same except that instead of `number` it's using `integer` pretty much everywhere (except for _one_ single place where there's still a `number` value! `Node.constnum`, which stores the resolved constant number value of a key of type `{ [1.5] = "foo" }` (we don't really care about non-integer keys... I could probably just change `Node.constnum` to `integer` and not store it for float keys; then `is_positive_int` could be simplified to `n and n >= 1`).

Converting the codebase to integers wasn't really that much work, especially given the editor integration powered by `tl types`, which makes it easy to see which side of the operation is turning the overall result into `number` when you get an `expected integer, got number`. (But I really missed having a proper "go to definition"!) — this makes me confident that switching over to an integer-enabled version shouldn't be too painful for users of the language.

One thing that is concerning me about the whole integer thing is adding another subtyping relation, and its consequences for things like `{number}` vs. `{integer}` (the same old variance issues crop up). Mixing subtyping, mutability, generics, polymorphism and inference is a recipe for trouble, but at this point we're already knee-deep in this mess. I feel like I'll have to introduce explicit variance annotations sooner rather than later.

hishamhm:
CI seems to be failing because Travis is hitting some rate-limiting on Github ¯\_(ツ)_/¯

hishamhm:
I don't have a whole lot of Teal code I can test changes other than the compiler and the test suite, but I just tested this branch on the [cyan](https://github.com/teal-language/cyan) codebase to see the amount of breakage it might cause, and I'm happy to report that it just reported two errors, requiring a grand total of 4 lines changed:

```
diff --git a/src/cyan/graph.tl b/src/cyan/graph.tl
index c0dec89..78584b8 100644
--- a/src/cyan/graph.tl
+++ b/src/cyan/graph.tl
@@ -114,8 +114,8 @@ function graph.scan_dir(dir: string | fs.Path, include: {string}, exclude: {stri
       end
    end

-   local cache <const>: {Node:number} = {}
-   local function count_deps(n: Node): number
+   local cache <const>: {Node:integer} = {}
+   local function count_deps(n: Node): integer
       if cache[n] then return cache[n] end
       local deps = 0
       for _, v in ipairs(n.dependents) do
diff --git a/src/cyan/tlcommon.tl b/src/cyan/tlcommon.tl
index 8c8c593..7b3d0ce 100644
--- a/src/cyan/tlcommon.tl
+++ b/src/cyan/tlcommon.tl
@@ -19,8 +19,8 @@ local type Node = any -- tl.Node
 
 -- tl.Token also not exposed :((((((((
 local type Token = record
-   x: number
-   y: number
+   x: integer
+   y: integer
    tk: string
    kind: string
 end
```

@euclidianAce ^

If anyone else has Teal code of their own that they could test with this branch, it would be much appreciated!

euclidianAce:
just doing some small tests everything seems to be in working order. One question I have is if arrays are going to change from `{number:T}` to `{integer:T}`? I assume this would break some compatability, but if people are going through codebases and changing a bunch of `number` to `integer` I don't think this would be too bad

hishamhm:
> One question I have is if arrays are going to change from {number:T} to {integer:T}?

Ohh, good question! The current iteration of the PR (as of 4cc8500) does not change that, and `t[x]` given a `number` x doesn't fail (and, as usual, will just return `nil` given an invalid key to the array).

I just changed it locally here to make array indices be integers and see the impact. The only places where it complained for `tl.tl` were places where `constnum` (now pretty much the only `number` in the code) was being used to index the `types[]` array of tuples. In those cases they were not bugs because there were either checks ahead of those accesses to certify that the index was integer, or the code was written so that any invalid index would produce a nil value which would be subsequently handled. But those checks could have been missed! So, my current line of thought is that there is value to make array indices to be integers... 

I re-checked cyan (with the 4-line diff above) using this modified compiler and the additional constraint that array indices must be integers did not trigger any additional compilation errors.

catwell:
I have updated my AoC code to support latest master (and found [a bug](https://github.com/catwell/adventofcode/commit/0a9533f949a3be7668073a74fb7f5645cec9c637) by the way, not sure if it has been reported) the tried to run it using this branch and got issues.

For instance I got this:

```
2 errors:
../util.tl:99:30: in assignment: got number, expected integer
../util.tl:105:30: in assignment: got number, expected integer
```

where the code in question is [this](https://github.com/catwell/adventofcode/blob/0a9533f949a3be7668073a74fb7f5645cec9c637/2020/util.tl#L97-L107). It's all numbers here, there's no reason to expect an integer.

I also got a lot of more understandable breakage where the compiler doesn't like me indexing arrays with "numbers", and errors like this:

```
./common.tl:46:9: in assignment: got number, expected integer
./common.tl:46:14: in assignment: got number, expected integer
```

on [this line](https://github.com/catwell/adventofcode/blob/0a9533f949a3be7668073a74fb7f5645cec9c637/2020/day03/common.tl#L46).

All in all, *all* numbers I used were integers so the port is straightforward and looks [like this](https://github.com/catwell/adventofcode/commit/b0b23dadc9c33b7f3ff9d86defce773e08d30701).

hishamhm:
> I understand why it happens but it still breaks existing code and looks like a bug

I'm okay with breaking existing code, but I'd rather it not looked like a bug. I wonder if something like `got number, expected integer (inferred at util.tl:98:8)` would make it more immediately clear?

> (until you realize that you have to explicitly cast 0 and 1 as numbers...)

both `local r = 0.0` or `local r: number = 0` would do the trick!

> All in all, all numbers I used were integers so the port is straightforward and looks like this.

Yeah, it sounds reasonable to me to require those arguments to be changed to integers.

Thank you for the feedback! It's super valuable!

lenscas:
tried it on my own project,  only 2 errors got reported before I actually changed my .d.tl files. So, it seems to work with minimal breakages.

So, I guess it is time to update tealr :)

hishamhm:
> I have updated my AoC code to support latest master (and found a regression by the way, not sure if it has been reported, but it's not very severe given you can do this)

Thanks for pointing it out! Fixed [here](https://github.com/teal-language/tl/commit/cc76ac6f2a4575a4c95888a3e2982e598db22e55)!

hishamhm:
All right, I'm confident now about merging this! Thank you all for the testing and feedback for the last 15 days! :heart: This is an important step forward :) :rocket:

PR #408, opened 2021-02-27
fix: propagate failstore in an incomplete statement

hishamhm:
@pdesaulniers This should fix the issues you were observing it `tl types`


pdesaulniers:
Yep, `by_pos` now contains more data on partial input :+1: 

```lua
require("os").
```

```json
    "by_pos": {
        "test.tl": {
            "1": {
                "1": 1,
                "8": 4,
                "9": 3,
                "14": 4
            }
        }
    }
```

PR #414, opened 2021-03-06
tl.tl API cleanups and tl refactors

hishamhm:
## tl.tl API changes

* make unknowns a kind of warning — this means they can be silenced or turned into errors via the regular methods for warnings in the CLI
* only ever return `Result` values, as their names imply, do not take them as inputs
* each `Result` only contains errors for the given file
  * use `env.loaded` (and `env.loaded_order`) or `result.dependencies` to traverse all results and get all errors, as seen in the `tl` refactor below.

## tl refactors

They're not as pretty as @euclidianAce's refactors in past recent PR, but this is more of a preparation for the deprecation of `tl build` and to use the API changes above.

* move everything related to "tl build" to its own "submodule", preparing for deprecation ([Cyan](https://github.com/teal-language/cyan) should be used as a build system instead)
* moves all "top-level" code to the end
* secondary quality-of-life changes while I was at it:
   * always look for tlconfig.lua in parent dirs (and use a dumber but faster method)
   * `tl types`: emit JSON map keys in alphabetical order — the order shouldn't matter, but it should make it easier to read when debugging


euclidianAce:
For what its worth (as I havent looked at the code much) I have a local branch of both cyan and teal-language-server up and running with this new api and they are both (seemingly) working. (And if anything isn't I'd blame that on my own test coverage rather than these changes)

hishamhm:
@euclidianAce Thank you for this feedback, it's super valuable.

I'll go ahead and merge this. Another thing I will add is to move the version string from the `tl` script into the API, so it can be checked by clients. :+1:

PR #415, opened 2021-03-08
Nested records

Zash:
Fixes that `tl check test.tl` on a file with this:

```lua
local record foo
   record bar
   end
   { bar }
end
```

would report:

> test.tl:4:4: unknown type bar

Change suggested by @hishamhm


hishamhm:
@Zash thank you!

PR #420, opened 2021-03-15
metamethod support for __tostring, __pairs, __gc

kikito:
This seems like an omission in #299. I couldn't find an obvious reason why it wasn't included on that PR. But the tests seem happy, so here it is :)

Related with #224

### DIFF
--- a/spec/metamethods/tostring_spec.lua
+++ b/spec/metamethods/tostring_spec.lua
@@ -1,0 +1,36 @@
+local util = require("spec.util")
+
+describe("unary metamethod __tostring", function()
+   it("can be set on a record", util.check [[
+      local type Rec = record
+         x: number
+         metamethod __tostring: function(Rec): string
+      end
+
+      local rec_mt: metatable<Rec>
+      rec_mt = {
+         __tostring = function(self: Rec): string
+            return "Rec {x: " .. tostring(self.x) .. "}"
+         end,
+      }
+
+      local r = setmetatable({ x = 10 } as Rec, rec_mt)
+      print(r)
+   ]])
+
+   it("can be used on a record prototype", util.check [[
+      local record A
+         value: number
+         metamethod __tostring: function(A): string
+      end
+      local A_mt: metatable<A>
+      A_mt = {
+         __tostring = function(a: A): string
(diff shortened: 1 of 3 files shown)


hishamhm:
@kikito Interesting question!

I think the reason why that list did not contain `__tostring` is because `__tostring` and `__pairs` are metamethods "at the standard library level" (they are implemented by `tostring()` and `pairs()`) and not "at the language level" (there is actually core VM code to handle the other ones, even if you don't `luaopen_base()` when initializing a `lua_State`).

In Teal, this translates to the language needing to be aware of the core metamethods to allow expressions such as `e1 * e2` where `e1` is a record that implements `__mul`, for example. For `__tostring`, if you don't declare the `metamethod` in the `record` and just put `__tostring` in the metatable, the resulting code just works, because `tostring()` is declared as taking type `any` in its input. So we added explicit `metamethod` declarations _in the type_ to make the compiler statically aware of them (because the metatable is set dynamically, e.g. it can be inside an `if` and then the compiler can't be sure).

But the standard library definitions in Teal _are_ aware of `__tostring` and `__pairs` (so, for example, the compiler already complains if you define the `__tostring` callback with too many arguments). So I guess accepting their declaration in the type as well at minimum wouldn't hurt and would probably be desirable for documentation and to avoid surprises.

Do you think we should add `__gc` (that's an interesting one! it's VM-level but its behavior is pretty much invisible, so it was happily ignored by Teal) and `__pairs` as well to accepted metamethods name in type definitions as well? It sounds reasonable to me!

hishamhm:
(ah, small PS: please include the generated `tl.lua` in the PR as well! simply running `make` will update it)

hishamhm:
@kikito Do you want to add `__pairs` and `__gc` to this PR? (no need to add tests with use examples like you did for `__tostring`, declaration only would be fine). If not, could you run `tl gen tl.tl` and `git add tl.lua` to the PR so that it includes `tl.lua` and is mergeable? Thanks!

kikito:
@hishamhm done! I've also updated the pr name.

Thank you for your patience with this.

hishamhm:
@kikito thanks! don't worry, there's no hurry :)

PR #423, opened 2021-03-19
docs: Fix typo global_{dev,env}_def

Zash:
This confused me for a moment. I'm assuming that it's a typo since this variant isn't mentioned anywhere else, especially not the line above and the table of options.


hishamhm:
oops! Yes, it was a typo, thank you!

PR #424, opened 2021-03-19
Fix tlconfig def env

Zash:
`global_env_def` in `tlconfig.lua` had no effect, but the `--global-env-def` option worked. Found that the value from the config file was getting overwritten with the first item from the command line argument, even if there were zero such arguments.

Test case copied and modified version of the first test case in that spec file.

Curious that `args[]` fields end up with empty arrays like this.

Alternate solution could be to set `:count("?")`, but this behaves slightly differently.


hishamhm:
Thank you! I'm going to push a 0.13.1 release with this!

Issue #340, opened 2021-01-16
error messages do not print `...` for vararg arguments and returns in function signatures


hishamhm:
Error messages that print function types report incorrect types without the proper `...`

To reproduce:

```lua
local function f(x: number, ...: string): string...
   return ...
end

f = 2
```

The place to look in the source code is the  `show_type_base` function.

PR #425, opened 2021-03-22
Error messages now report proper `...`

SovietKitsune:
This PR tweaks the output of `show_type_base` to fix issue #340.

Before:
```lua
function(number, string): string
```

After:
```lua
function(number, string...): string...
```

### DIFF
--- a/spec/varargs/type_base_spec.lua
+++ b/spec/varargs/type_base_spec.lua
@@ -8,6 +8,6 @@ describe("type base", function()
 
       f = 2
    ]], {
-      { msg = "in assignment: got integer, expected function(number, string...): string..." }
+      { msg = "in assignment: got integer, expected function(number, ...: string): string..." }
    }))
 end)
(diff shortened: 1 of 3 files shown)


hishamhm:
Yay, thanks for the PR!

A quick heads up that this type declaration currently doesn't work in Teal:

```
local f: function(integer, string...): string...
```

but this does (because variable names are optional in type declarations)

```
local f: function(integer, ...: string): string...
```

(I also just noticed that `parse_argument_list` checks that `...` is only used at the last position, but `parse_argument_type_list` doesn't! that's another issue!)

I know that the displayed types in error messages don't necessarily match the `type` syntax in the language (anonymous records, etc.), but I think that in this case it might end up being extra confusing for users if the compiler displays `function(string...)` but the language doesn't accept it.

SovietKitsune:
Alright I fixed the syntax to be
```lua
function(number, ...: string): string...
```

hishamhm:
@SovietKitsune Merged, thank you!

PR #437, opened 2021-04-02
fix: allow # to be used on tuples

euclidianAce:
Apparently forgot to add tupletables to the `unop_types` table, this fixes that and adds a small sanity check for it


hishamhm:
merged! thanks for adding the tests :)

Issue #436, opened 2021-04-01
Should "error()" type definition be updated to match Lua manual?


JLPLabs:
Today in tl.tl we see `error()` defined as...
` ["error"] = a_type { typename = "function", args = TUPLE { STRING, NUMBER }, rets = TUPLE {} },`

This definition does meet the common usage, but it does not agree with the Lua manual.
Sect 2.3 has this paragraph... (emphasis added)

> Whenever there is an error, an error object is propagated with information about the error. Lua itself only generates errors whose error object is a string, but programs may generate errors with **any value** as the error object. It is up to the Lua program or its host to handle such error objects. For historical reasons, an error object is often called an error message, even though it **does not have to be a string**. 

An example demonstrating this appears below, as a postscript.

I think it would be more correct to accept any type; *HOWEVER* I have no idea what the implications are for other aspects of teal if such a fundamental change would be made.

So, this change *might* be in order:
` ["error"] = a_type { typename = "function", args = TUPLE { ANY, NUMBER }, rets = TUPLE {} },`

----
EXAMPLE of non-string 1st argument for `error()`:

    test = function (x)
      print("in test(); got x: "..x..". And now going to generate an error")
      error({code=123})
    end
    ok, res = pcall(test, 42)   --> in test(); got x: 42. And now going to generate an error
    ok                          --> false
    res.code                    --> 123

hishamhm:
Fixed by your own #438, thank you for the initiative!! :)

PR #438, opened 2021-04-02
Allows error() arg1 to be anything, not just a string.

JLPLabs:
Addresses Issue #436.

I can supply an example file demonstrating that with the fix error() will accept a table as arg1.


hishamhm:
@JLPLabs thank you! merged!

Issue #443, opened 2021-04-18
os.time signature problem?


remdragon:
teal seems to be telling me that os.time() doesn't take any arguments:

```
local now_utc = os.date ( "!*t" )
local epoch = os.time ( now_utc )
print ( epoch )
```

lua5.1 produces:
```
1618783034
```

tl check produces:
```
========================================
1 error:
tl_os_time_takes_arguments.tl:2:23: wrong number of arguments (given 1, expects 0)
```

hishamhm:
Yes, I just took a peek and there's a TODO comment in tl.tl for the definition to be fixed, shouldn't be difficult to fix to match the other definitions!

remdragon:
If it helps, the following gives me a warning about redefining os but otherwise passes and the OsDate record I put here is what I'm getting from the os.date() call and matches what I'm feeding as the parameter to os.time()
```
local record OsDate
	hour: integer
	min: integer
	wday: integer
	day: integer
	month: integer
	year: integer
	sec: integer
	yday: integer
	isdst: boolean
end
local record OS
	date: function ( unknown: string ): OsDate
	time: function ( unknown: OsDate ): integer
end
local os: OS
local now_utc = os.date ( "!*t" )
local epoch = os.time ( now_utc )
print ( epoch )
```

PR #445, opened 2021-04-20
teal now accepts table as input to os.time

JLPLabs:
This addresses #443.
In tl.tl used os.date as template; now os.time fulfills the Lua 5.4 manual:
os.time([table])


hishamhm:
@JLPLabs Thank you!

PR #446, opened 2021-04-21
Catch implicit redeclarations

euclidianAce:
This moves the check for redeclarations into `add_var` so implicit declarations in things like for loops and function arguments get caught


hishamhm:
@euclidianAce Nice! merged! :)

PR #453, opened 2021-06-19
Adding support for record + __call in forin

fcr--:
So, I realized this was a problem when I was writing the type definition for [luafun](https://github.com/luafun/luafun/blob/master/fun.lua#L47-L71) which in many cases it returns "iterable" callable records that have a ton of utility functions.

I'm not very familiar with teal's code, so I wouldn't be surprised if I'm missing something or if there's a better way to fix this.  However I tested this change and it seems to be working.

### DIFF
--- a/spec/statement/forin_spec.lua
+++ b/spec/statement/forin_spec.lua
@@ -89,6 +89,26 @@ describe("forin", function()
       end
    ]])
 
+   it("with a callable record interator", util.check [[
+      local record R
+         metamethod __call: function(): integer
+      end
+
+      function foo(): R
+         local x = 1
+         return setmetatable({}, {
+            __call = function(): integer
+               x = x + 1
+               return x < 4 and x or nil
+            end
+         })
+      end
+
+      for i in foo() do
+         print(i + 1)
+      end
+   ]])
+
    it("catches when too many values are passed", util.check_type_error([[
       local function it(): function(): string
          return nil
--- a/spec/statement/forin_spec.lua
+++ b/spec/statement/forin_spec.lua
@@ -91,23 +91,61 @@ describe("forin", function()
 
    it("with a callable record interator", util.check [[
       local record R
+         incr: integer
          metamethod __call: function(): integer
       end
 
-      function foo(): R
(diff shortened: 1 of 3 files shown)


hishamhm:
Thanks for the patch!

I took a look at tl.tl too see where else `__call` support was being implemented, to see if there was any possibility for code reuse, and I found `resolve_for_call`. This function is currently "local" to a `do` block for use by `type_check_function_call` but it looks like it could be moved out of that block to be reused by the `forin` visitor, in lieu of `resolve_tuple_and_nominal` (as `resolve_for_call` wraps that function with extra functionality including `__call`). At first I wasn't even sure I would mention this in this PR since it would be just a refactoring improvement, but then I noticed that function does something else as well: it adjusts the arguments for the function! 

In your test, your metamethod being called does not have arguments, but given a record r of type R with metamethod __call which is called as `r(a, b)`, `__call`is of type `function(R, A, B)` (ie, containing the record type for "self" and then whatever types to be used in the call). Lua auto-adds the `r` as a "self" into the metamethod callback, so we do the same in `resolve_for_call` [here](https://github.com/teal-language/tl/blob/master/tl.tl#L6622). So I think we actually need to do this refactor and reuse the function instead, to ensure __call metamethods with arguments are handled properly! 

I hope this description was not too confusing, let me know if it's clear enough!

fcr--:
Thank you very much for the detailed explanation!  I moved `resolve_for_call` outside `type_check_function_call` in order to reuse it.  However since I'm still not very familiar with the codebase I was not confident enough to try to typecheck the arguments of the iterator, so instead I added a TODO comment.  Hope that even being incomplete, this PR somehow helps!

hishamhm:
I fetched this branch locally and played a bit with it. I think I was misunderstanding argument checking — the TODO you put in is on point, and I think it already applied to existing iterator functions: only their returns are being used to determine the types of iterator variables, I don't think arguments are really checked. So, I think this PR is an improvement which is on par with the rest of the for-in functionality, so let's merge it as is and keep the TODO around! Thank you for revising the PR!

PR #461, opened 2021-07-07
Correct table.concat arg type to {string | number} (Fixes #450)

### LINKED ISSUE #450: table.concat expects {string} instead of {string|number}
Lua's reference manual [says](https://www.lua.org/manual/5.3/manual.html#pdf-table.concat)
> Given a list where all elements are strings or numbers, returns the string...

However,
```lua
local vector = {1,2,3}
print(table.concat(vector, ','))
```
causes Teal to error with
```
argument 1: got {integer}, expected {string}
```

Jcd1230:
Fixes #450

Started looking into this project, looks awesome for what I want to do with Lua (after spending way too much time looking into something statically typed to replace/augment Lua with!)

Decided to try fixing this to let myself get a better idea of the language, seemed like an easy enough fix. 

Let me know if the style isn't what you would expect, I wasn't exactly sure if it would be cleaner to just add the type inline where it was needed since its only used once. (Also since ARRAY_OF_STRING is only used once now as well)

Thanks for making Teal & supporting the Lua community!

### DIFF
--- a/spec/stdlib/table_spec.lua
+++ b/spec/stdlib/table_spec.lua
@@ -12,4 +12,11 @@ describe("table", function()
       ]])
    end)
 
+   describe("concat", function()
+      it("can concat a mixed list of strings and numbers", util.check [[
+         local t = { "a", 1, "b", 2 }
+         print(table.concat(t))
+      ]])
+   end)
+
 end)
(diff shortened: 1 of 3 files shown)


hishamhm:
The fix looks good! Thank you for writing the patch and including a test! It's now merged and will feature in the next release.

PR #473, opened 2021-09-01
compat: search for luajit's `bit` module after `bit32` not found

euclidianAce:


hishamhm:
I've avoided luajit's bit module before because it has different semantics
(those differences were the source of long discussions in the Lua mailing
list), so depending on bit32 (which can be loaded into luajit as a rock
like compat53) yields more consistent behavior for Teal as a language.

euclidianAce:
Ah, fair enough :+1:

Issue #97, opened 2020-04-30
Standardized class model?


pdesaulniers:
This has been mentioned in https://github.com/teal-language/tl/issues/86#issuecomment-611241592.

I think it's a good idea (although Lua purists might not like it?)

But then, which features should Teal support? Inheritance? Access modifiers? Getter-setter properties? Interfaces?

hishamhm:
This is a big question!

> I think it's a good idea (although Lua purists might not like it?)

I tend to lean towards "good idea" but I also realize that it could turn a lot of people off to the language (and maybe another lot of people on?).

The reason why I think about it is not so much to get the full kitchen sink of OOP features into the language, but for two main reasons:

* the #1 reason to me is to give users a statically-verifiable means for dealing with metatables. By providing something that does the metatable plugging, we avoid homegrown solutions and give control for that to the Teal compiler, which can then make better assumption about the behavior of metatable-infused objects. BUT — I wonder if we could just do that by adding metatable-specifics to `record` definitions. You can already model methods and attributes with them, so they're almost objects. We just don't have a neat way to do inheritance, to get the OOP baseline down.
* it's my understanding (hope I'm not getting the history backwards — @mascarenhas, I hear you're the TS expert now :) , do you know?) that TypeScript adding classes helped to standardize the JavaScript class model. So perhaps by pushing one model, Teal could help promote "Teal-compatible classes" (not in syntax but in behavior) even in pure Lua code. BUT — that's bluesky wishful thinking and we don't have the strength to make that push at this point. Anyway, I don't think one of Teal's goals should be to push Lua one way or another; it could be just a nice side-effect.

I remember @mascarenhas suggested once adding a class model to Typed Lua or Titan with "pluggable backends" so that they could emit code for different Lua OOP models — while the idea is intriguing I think it is overkill for Teal.

mascarenhas:
TypeScript did get classes years before ES6 was finalized and classes officially got into JavaScript, but they had the fig leaf that the proposal was already going through standardization, ES6 just took a loooong time. :smile: 

I have always been in favor of baking in classes and interfaces in the language definition, and Typed Lua would have them if we kept going; all the alternatives PL people have come up with to model the same kinds of concepts that OO models are more complex. :smile:

I was looking at the docs, does this work, and what does it compile to:

```
local Point = record
   x: number
   y: number
end

function Point:move(dx: number, dy: number)
   self.x = self.x + dx
   self.y = self.y + dy
end

[...]

PR #474, opened 2021-09-06
Record embed and const field syntax

pigpigyyy:
Implement the "embed" and "const / readonly" field functions for OOP simulations to support some Lua project uses. The "embed" function is refactored from @euclidianAce's [implementation](https://github.com/euclidianAce/tl/tree/record-embed) ( mentioned in issue #97 ) with bug fixes.

### DIFF
--- a/spec/declaration/record_spec.lua
+++ b/spec/declaration/record_spec.lua
@@ -673,3 +673,91 @@ for i, name in ipairs({"records", "arrayrecords"}) do
       ]]))
    end)
 end
+
+describe("embedding", function()
+   it("embed is not a reserved word", util.check [[
+      local record A
+         embed: number
+      end
+   ]])
+   it("should make members of a record available to another", util.check [[
+      local record A
+         userdata
+         a_value: number
+      end
+      local record B
+         embed A
+         b_value: number
+      end
+      local record C
+         embed B
+         c_value: string
+      end
+      local b: B = {}
+      local c: C = {}
+      print(b.a_value, c.a_value + c.b_value, c.c_value)
+      local function f(a: A, b: B) end
+      f(b, b)
+      f(c, c)
+      local a: A = b
+      a = c
+   ]])
+   it("should allow for generics", util.check [[
+      local record A<T>
+         property_of_a: T
+      end
+      local record B<T>
+         embed A<T>
+         property_of_b: T
+      end
+      local record C<T>
+         embed B<T>
+         property_of_c: string
+      end
+      local b: B<number> = {}
+      local c: C<number> = {}
+      print(b.property_of_a + 1, c.property_of_b + c.property_of_a, c.property_of_c)
+      local function f(a: A<number>, b: B<number>) end
+      f(b, b)
+      f(c, c)
+      local a: A<number> = b
+      a = c
+   ]])
+   it("embed multiple records", util.check [[
+      local record A
+         x: number
+      end
+      local record B
+         y: string
+      end
+      local record C
+         embed A
+         embed B
+         z: boolean
+      end
+      local c: C = { x = 1, y = "hello", z = true }
+      print(c.x, c.y, c.z)
+      local function f(a: A, b: B, c: C) end
+      f(c, c, c)
+   ]])
+   it("share the same element type with embeded arrayrecord", util.check [[
+      local record A
+         { string }
+         x: number
+      end
+      local record B
+         embed A
+         y: string
+      end
+      local record C
+         embed B
+         z: boolean
+      end
+      local c: C = { x = 1, y = "hello", z = true }
+      local str: string = c[1]
+      print(c.x, c.y, c.z)
+      local strs: {string} = c
+      local b: B = c
+      strs = b
+   ]])
+end)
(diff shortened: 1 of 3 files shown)


hishamhm:
Hi @pigpigyyy, thank you for taking the time to make a PR!

The code itself looks good, but I'm uncertain about adding these features to the language. These are the reasons:

* The `embed` feature seems to suggest a structural subtyping system, but Teal records are nominal. I'm thinking of adding explicit subtyping, but that needs cleaning up the logic for covariant/contravariant checks — since Lua tables are used in both modes, there would need to be some annotation for covariant/contravariant arguments, and making this in a non-confusing manner for using is an interesting design problem.

* The readonly feature for record fields is interesting, but I'm afraid it might provide a false sense of security, since fields could be rewritten by simply aliasing the record to a different type...

I'd like to hear what others think about these concerns!

lenscas:
> The readonly feature for record fields is interesting, but I'm afraid it might provide a false sense of security, since fields could be rewritten by simply aliasing the record to a different type...

In most languages there are also ways to get around fields being public, private and/or protected. Yet those languages don't give that idea up and properly setting the access for fields is encouraged instead of having everything be public.

Also, by that same logic, the entire type system gives a false sense of security as a simple `as` can be used to give a table to a function that takes an integer.

I think it is best to treat `as` as how Rust treats `Transmute`, it is there if you really need/want it but... YOU need to make sure you don't screw up.

pigpigyyy:
I'm using Teal in a C++ project with Lua embedded, and get a lot of native OOP stuff exported in Lua. So I just seek for a way to do the simulated OOP check. Since C++ does not support covariant/contravariant features by default, so I'm not worrying too much for it. As for nominal records, I just changed the way to compare records with embeds in order to make subtypes to be not equal.

hishamhm:
> I think it is best to treat as as how Rust treats Transmute, it is there if you really need/want it but... YOU need to make sure you don't screw up.

@lenscas What I meant by aliasing is different from an explicit `as` because there are places where we have bivariant behavior (due to lack of proper support for record subtyping yet) so that causes implicit aliasing. I agree that in explicit `as`, all bets are off. My concern was about implicit aliasing and changing those rules in the future, hope that clarifies!

hishamhm:
> As for nominal records, I just changed the way to compare records with embeds in order to make subtypes to be not equal.

Thanks @pigpigyyy, I'll give it another review!

lenscas:
> > I think it is best to treat as as how Rust treats Transmute, it is there if you really need/want it but... YOU need to make sure you don't screw up.
> 
> @lenscas What I meant by aliasing is different from an explicit `as` because there are places where we have bivariant behavior (due to lack of proper support for record subtyping yet) so that causes implicit aliasing. I agree that in explicit `as`, all bets are off. My concern was about implicit aliasing and changing those rules in the future, hope that clarifies!


Ah, yes then I fully agree to wait until that to see if `readonly` is a good fit.

muratg:
@hishamhm Did you get a chance to look into this a bit more?  I'm personally loving the "embed" design and implementation here. It's feels almost like intersection types from TypeScript. 

I'm not worried type-soundness, but being able to express another much-common usage of tables in the type system. 

I'm considering taking this PR in a temporary fork if you're likely get this in or something similar in the future.

(I'm currently using TypescriptToLua but taking a serious look at Teal.)

hishamhm:
@muratg Thanks for the feedback!

Regarding this feature proposal, I think the direction Teal should go is to adopt real intersection types. Various features in the language already hint in this direction (arrayrecords, polymorphic function members in records) so it would be cleaner to replace them with intersection types. I'm sure that intersections would end up having some limitations similar to those we currently have with unions, but I think overall it would make the design cleaner (and the language hopefully smaller, which is not a hard goal but always a hint of elegance).

Regarding the other syntax changes, yeah, no intention on changing them now. (Side comment: it's always interesting to see people's different perceptions... most of your suggestions lean towards making it look more like Typescript or Go, which is probably what you're most used to! "Intuitive" is always relative to the beholder — Teal's syntax style follows on the Lua tradition, which is in itself more aligned to the Pascal/Modula family than that of curly-bracket-style languages).

muratg:
I agree familiarity is important in deciding what's intuitive.  My main issue with colon in map and function syntax was more about `:` being overused. In the typing context, a colon usually comes after a variable name or a function argument.  But for map it's coming after another type, for example. There's a similar issue in Typescript too. But in Lua/Teal syntax it's a slightly bigger issue, because `:` is also used to denote "methods", both in definition and use site. 

But I can totally see your point of view too.  It makes sense to target existing Lua users for Teal instead of other languages.

Looking forward to seeing the intersection types!  I think it should have less limitations compared to union types since it's an extension process and not a contraction.  If you already thought of its syntax, please share!

hishamhm:
I'm closing this PR since we'll be getting functionality to address similar use-cases via the `interface` type as already implemented in the `next` branch! Thank you all for the brainstorming and feedback!

Interfaces still need some work (generics, ensuring the subtyping relations all work as intended (tests!!!!)) but I'm pretty happy with the implementation to the point that I've already ported the compiler's own type objects to use them :)

pigpigyyy:
Just started testing the Next Teal! Everything is supposed to be there except the "readonly field" feature, I will try making a new PR with the newest version. Thanks for your effort!

PR #485, opened 2021-09-27
Fix a case to check array of enums

pigpigyyy:
An array of enums used as argument in a overloaded function is not well checked. A minimal example could be
```lua
local enum E
   "a"
end
local record R
   f: function()
   f: function({E})
end
local r: R
r.f({"a", "ERROR"}) -- is not reporting type errors
```

### DIFF
--- a/spec/declaration/record_spec.lua
+++ b/spec/declaration/record_spec.lua
@@ -129,6 +129,22 @@ for i, name in ipairs({"records", "arrayrecords"}) do
          { msg = "attempt to redeclare field 'print' (only functions can be overloaded)" }
       }))
 
+      it("enum check in overloaded function", util.check_type_error([[
+         local enum E
+            "a"
+            "b"
+            "c"
+         end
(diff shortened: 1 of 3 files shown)


hishamhm:
Thanks for the fix!

An interesting thing is that this uses the internal `types` attribute from arrays of string literals, which was added for implementing tuple-arrays. Nice reuse there.

String literals themselves are a second-class type in Teal: the type of a string literal carries knowledge about the literal itself in order to make literal-to-enum comparisons, but you can't declare a variable as a literal. This fix has an implication to the underlying type theory that array-of-string-literals also becomes a second-class type, implementing its intuitive behavior.

Speaking of second-class types: like Lua, I don't think it's necessarily a bad thing to have second-class types (`...` is second-class in Lua). I don't want to endlessly grow the typesystem a la TypeScript, so something being second-class is not necessarily a problem to be fixed. Having said that, we might add first-class string literal types in the future, since I've seen (and used!) single-element enums to simulate it.

PR #488, opened 2021-10-07
make math.huge an integer instead of a number

kikito:
Otherwise initializing an integer variable to math.huge throws an error:

```
local x: integer = math.huge -- expected integer, got number
```

### DIFF
--- a/spec/stdlib/math_spec.lua
+++ b/spec/stdlib/math_spec.lua
@@ -16,4 +16,10 @@ describe("math", function()
       ]])
    end)
 
+   describe("huge", function()
+      it("is an integer", util.check [[
+         local x: integer = math.huge
+      ]])
+   end)
+
 end)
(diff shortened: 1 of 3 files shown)


hishamhm:
I think the best alternative to this PR would be instead to make the compiler auto-generate that `math.maxinteger` line as needed, similarly to how it handles the `table.unpack` code, ensuring that it works even when using `--gen-compat=off`. Most of the machinery to handle it is already there.

Relevant bits:

* https://github.com/teal-language/tl/blob/b8ee3cba9dec1dca1272f020535ec478182b4dd3/tl.tl#L4565-L4566 - generating the code
* https://github.com/teal-language/tl/blob/b8ee3cba9dec1dca1272f020535ec478182b4dd3/tl.tl#L5138 - marking the entry in the standard library defs with `needs_compat = true`

Issue #521, opened 2022-05-20
Failure → spec/loader/loader_spec.lua @ 7:  tl.loader on .tl files reports filename correctly in debug info


Koeng101:
On a fresh git clone, I tried the following test:

```
luarocks test

tl dev-1 depends on dkjson (2.6-1 installed)
[...]

PR #522, opened 2022-05-20
Fix Lua 5.1 error (load -> loadstring)

Koeng101:
I was attempting to run teal within gopher-lua - https://github.com/yuin/gopher-lua - and was running into a few odd errors. I was running the following scripts:

```go
//main.go
package main

import (
        "github.com/yuin/gopher-lua"
)

var luaString string = `
tl = require("tl")
tl.loader()
addsub = require("addsub")
print (addsub.add(10, 20))
`

func main() {
        L := lua.NewState()
        defer L.Close()
        if err := L.DoString(luaString); err != nil {
                panic(err)
        }

}
```

```teal
--addsub.tl
local addsub = {}

function addsub.add(a: number, b: number): number
   return a + b
end

function addsub.sub(a: number, b: number): number
   return a - b
end

return addsub
```

The code block `i, node.decltype = i, decltype` would cause a bubbling up error (since all the lua types were checked by Go) where `i` would occasionally be set to "decltype" (I have no idea why).  The second edit is just a little check for 5.1, since Lua 5.1 doesn't have load, I added in loadstring if it detects that version. These 2 edits let this script run properly so that I can run teal in Golang, which is pretty neat.

Happy to fix up formatting or anything else - tonight is the first time I've ever used lua, so I'm not sure about conventions here.

Note: on my computer, not all tests pass. Made issue #521 about that. All the tests that pass in master repo also pass in this updated repo.

### DIFF
--- a/tl.lua
+++ b/tl.lua
@@ -2162,7 +2162,7 @@ local function parse_argument(ps, i)
       i, decltype = parse_type(ps, i)
 
       if node then
-         i, node.decltype = i, decltype
+         node.decltype = decltype
       end
    end
    return i, node, 0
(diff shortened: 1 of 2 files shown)


Koeng101:
Simplifying this PR a little bit.

PR #523, opened 2022-05-29
Fix redeclaring `i`

Koeng101:
When running teal in gopher-lua (https://github.com/yuin/gopher-lua), types get checked by the system, and the single line `i, node.decltype = i, decltype` caused a complete failure to run teal. When replaced with the equivalent `node.decltype = decltype`, the system runs just fine.

These should be functionally equivalent. Nothing else is changed and it does not appear to change any tests.

### DIFF
--- a/tl.lua
+++ b/tl.lua
@@ -2162,7 +2162,7 @@ local function parse_argument(ps, i)
       i, decltype = parse_type(ps, i)
 
       if node then
-         i, node.decltype = i, decltype
+         node.decltype = decltype
       end
    end
    return i, node, 0
@@ -9634,7 +9634,16 @@ local function tl_package_loader(module_name)
 
 
       local code = assert(tl.pretty_print_ast(program, tl.package_loader_env.gen_target, true))
-      local chunk, err = load(code, "@" .. found_filename, "t")
+
+
+      local err = ""
+      local chunk = function() end
+      if _VERSION == "Lua 5.1" then
+         chunk = loadstring(code)
+         err = "5.1 error"
+      else
+         chunk, err = load(code, "@" .. found_filename, "t")
+      end
       if chunk then
          return function()
             local ret = chunk()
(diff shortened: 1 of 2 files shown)


tooolbox:
Nice.

Refs:
- https://github.com/yuin/gopher-lua/issues/315
- https://github.com/yuin/gopher-lua/issues/314

tooolbox:
@Koeng101 I'm curious what you ended up doing for the `load` vs. `loadstring` bit, because I saw you threw in a fix for that and then took it out.  I'm using `compat52.lua` and its dependencies to run Teal in Gopher-Lua, but it's not ideal since it calls `string.dump` which blows up, so I had to modify the lib.

Koeng101:
@tooolbox I added in a check into tl.tl that if it detected version 5.1, it would use `load` instead. The link is here - https://github.com/teal-language/tl/pull/522/commits/4406b736d2b377faf1f35db2db1e3ded57019499

kinda dumb simple:
```lua
      if _VERSION == "Lua 5.1" then
         chunk = loadstring(code)
         err = "5.1 error"
      else
         chunk, err = load(code, "@" .. found_filename, "t")
      end
```

I threw it out because I thought it might not be idiomatic or well made lua code. For my application, I actually want to use the lua conversion, so I just do this:

```go
err = L.DoString(fmt.Sprintf("return tl.gen([========[%s]========])", tealCode))
        if err != nil {
                return "", err
        }
        parsedLua := L.Get(1)
```

It outputs a lua string, which I save, and then I just run the outputted lua. Kinda hacky, but is what I wanted to do in the first place.

I had no idea that https://github.com/yuin/gopher-lua/issues/314 was an issue, or else I probably would have been a lot less inclined to figure out something that worked.

tooolbox:
Gotcha, so you're loading the Lua string yourself.  Okay, thanks for sharing that.

PR #538, opened 2022-07-20
canonicalize pathnames to improve behavior of env.loaded

### LINKED ISSUE #518: Error with custom (non-native) types in global variable declarations in .d.tl files
```
-- somefile.d.tl:
local type Foobar = number -- Global or local same results, alias or record same results.
local MyLocalNumber: number -- Works
local MyLocalFoobar: Foobar -- Works
global MyGlobalNumber: number -- Works
global MyGlobalFoobar: Foobar -- Error: cannot redeclare global with a different type: previous type of MyGlobalFoobar is Foobar
-- Only generates the above error in .d.tl files, not in regular .tl files.
-- EDIT:
-- Despite the error, the declaration is still in effect for dependent .tl files (in my particular case this is a global_env_def).
-- The error won't show up when you run "tl build", but it shows up when you run "tl check somefile.d.tl"
-- It also shows up in vscode with the vscode-teal plugin, but again only for .d.tl files, not regular .tl files.
```

hishamhm:
This does not go as far as converting pathnames to absolute as that would require going beyond pure-Lua, but it does improve the situation. In particular, it should allow type-checking the global_env_def file from the current folder without a confusion between "file.d.tl" and "./file.d.tl".

Fixes #518 (unless relative and absolute paths are mixed, but it's an incremental improvement)

### DIFF
--- a/spec/cli/build_script_spec.lua
+++ b/spec/cli/build_script_spec.lua
@@ -282,7 +282,7 @@ Wrote: build/build.lua
          cmd_output =
 [[========================================
 1 syntax error:
-./build.tl:8:17: syntax error
+build.tl:8:17: syntax error
 ]]
       })
    end)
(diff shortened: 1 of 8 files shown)


hishamhm:
Closed in favor of #540

PR #543, opened 2022-08-03
Require "global function" in global functions

### LINKED ISSUE #500: Global functions
Tutorial says that one does not simply define a global variable without the `global` keyword and then it shows the example of such a function:

```lua
function test(...: number): number...
  return ...
end
```

... which defines global variable `test` without a `global` keyword.

hishamhm:
This PR changes the language so that bare `function name() ... end` doesn't work anymore: you need to say either `local function name() ... end` or `global function name() ... end`.

Record functions still work the same way:
* `function record:field() end` and `function record.field() end` still work.

Function values still work the same way:
* `x = function() end` still works.

Bare `function name() ... end` still works if the file extension is `.lua` instead of `.tl` (triggering the compiler's "lax mode").

Initial feedback regarding this change was positive back in February when I brought it up in the Matrix/Gitter chat in the wake of #500, but I'd still like to get feedback before merging this significant change!

Closes #500.

### DIFF
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -48,7 +48,7 @@ precedence, see below.
 
    label ::= ‘::’ Name ‘::’
 
-   funcname ::= Name {‘.’ Name} [‘:’ Name]
++  funcname ::= Name {‘.’ Name} ‘:’ Name | Name {‘.’ Name} ‘.’ Name
 
    varlist ::= var {‘,’ var}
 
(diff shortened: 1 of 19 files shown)


hishamhm:
@pdesaulniers @euclidianAce Not sure if this proposed change has any impact on tooling (any existing parsers you are using will probably continue working, but the compiler itself will produce additional syntax errors on bare `function` statements (technically, the error is being produced in the type-checking stage because I use the `lax` flag, but I think that has no difference in output)).

PR #555, opened 2022-08-26
Fix a few compiler issues with close attribute

pigpigyyy:
Fix a few issues when using Teal with close attributes feature. Stop vscode plugin from complaining compiler crashes.


hishamhm:
Cool, thank you! Pinging @euclidianAce for a quick look since he was the one worked on the close attribute support.

hishamhm:
I went ahead and merged this since it looks good to me :) — if there's anything else to be addressed in this topic, it can be done in future PRs!

PR #556, opened 2022-08-26
Generate cleaner codes for nil check

pigpigyyy:
Have been using an idiom to notice user to do a nil check before using some API results.
```lua
local record R
   f: function()
end
local function get(): R | nil
end
local r = get()
if not r is nil then
   r.f()
end
```
And currently `r is nil` is being generated to `(type(r) == "nil")`, could make it to be just `(r == nil)`.

### DIFF
--- a/spec/operator/is_spec.lua
+++ b/spec/operator/is_spec.lua
@@ -356,6 +356,27 @@ describe("flow analysis with is", function()
          { y = 10, msg = "got boolean | thread, expected boolean" },
       }))
 
+      it("gen cleaner checking codes for nil", util.gen([[
+         local record R
+            f: function()
+         end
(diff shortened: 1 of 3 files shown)


lenscas:
wouldn't this potentially break for tables/userdata that implement their own comparison? As there `a == nil` can mean something else than `type(a) == "nil"`

pigpigyyy:
The `__eq` metamethod will only be invoked when Lua find both side items from "==" expression to be tables or userdata.

lenscas:
Ok. Sounds weird but then I guess this indeed doesn't break anything/change behaviour.

hishamhm:
@pigpigyyy Nice, thank you!

PR #558, opened 2022-09-05
Fix alt text of gitter badge

XeCycle:
It pointed to coreclr, but we are tl.

Link is copied from bottom of the file.


hishamhm:
@XeCycle Merged, thank you for spotting and the PR!

PR #566, opened 2022-09-25
improve error message for redefined functions

catwell:
Consider this code sample:

```
local function foo() end
local function foo() end
```

Before this change:

```
3 warnings:
invalid.tl:2:1: redeclaration of variable 'local' (originally declared at 1:1)
invalid.tl:1:1: unused function foo: function()
invalid.tl:2:1: unused function foo: function()
```

After this change:

```
3 warnings:
invalid.tl:2:1: redeclaration of function 'foo' (originally declared at 1:1)
invalid.tl:1:1: unused function foo: function()
invalid.tl:2:1: unused function foo: function()
```


hishamhm:
good catch! thanks for the fix!

PR #567, opened 2022-09-25
Allow predeclaration of local functions

catwell:
Consider this code sample:

```
local foo : function()
function foo() end
foo()
```

Before this change it would fail with:

```
1 error:
invalid.tl:2:1: functions need an explicit 'local' or 'global' annotation
```

This change makes it valid again (it was the case before 0.14).

Rationale: in Lua `function foo() end` is sugar for `foo = function() end`, and I suppose it should remain the same in Teal.

This kind of code is useful to write mutually recursive functions such as this:

```
local collatz_odd : function(integer): integer

local function collatz_even(n: integer): integer
    print(n)
    n = n // 2
    if n % 2 == 1 then
        return collatz_odd(n)
    end
    return collatz_even(n)
end

function collatz_odd(n: integer): integer
    print(n)
    if n == 1 then
        return 1
    end
    return collatz_even(3 * n + 1)
end

collatz_odd(7)
```

### DIFF
--- a/spec/declaration/global_function_spec.lua
+++ b/spec/declaration/global_function_spec.lua
@@ -58,6 +58,13 @@ describe("global function", function()
       { y = 1, msg = "functions need an explicit 'local' or 'global' annotation" },
    }))
 
+   it("a function can be pre-declared", util.check [[
+      local f: function()
+      function f()
+         print("I am a bare function")
+      end
+   ]])
+
    local modes = {
       {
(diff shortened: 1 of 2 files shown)


hishamhm:
I had no strong feelings myself about supporting or not this Lua idiom, but I suppose it is consistent and unambiguous enough, and given the finished PR, I'm happy to merge it :)

PR #570, opened 2022-10-09
Skip UTF-8 BOM when reading file (#435)

### LINKED ISSUE #435: Unicode BOM symbol support.
Hi. I do not know is it real issue, but then my editor save BOM symbols in file(on Windows) Teal parser get error like this:
```
3 syntax errors:
sub\code2.tl:1:1: invalid token '∩'
sub\code2.tl:1:2: invalid token '╗'
sub\code2.tl:1:3: invalid token '┐'
```

sveyret:
Closes #435.

### DIFF
--- a/spec/api/process_spec.lua
+++ b/spec/api/process_spec.lua
@@ -36,5 +36,25 @@ describe("tl.process", function()
          assert(foo_result, foo_err)
          assert(bar_result, bar_err)
       end)
+      it("should strip BOM from files", function()
+
+         local bom = "\xEF\xBB\xBF"
+         local current_dir = lfs.currentdir()
+         local dir_name = util.write_tmp_dir(finally, {
+            ["main.tl"] = bom .. [[
+               return "working"
+            ]],
+         })
+         local expected_lua_code = [[
+            return "working"
+         ]]
+
+         assert(lfs.chdir(dir_name))
+         local result, err = tl.process("main.tl")
+         assert(lfs.chdir(current_dir))
+         assert.same({}, result.syntax_errors)
+         local pretty_result_string = tl.pretty_print_ast(result.ast)
+         util.assert_line_by_line(expected_lua_code, pretty_result_string)
+      end)
    end)
 end)
--- a/tl.lua
+++ b/tl.lua
@@ -9850,6 +9850,15 @@ end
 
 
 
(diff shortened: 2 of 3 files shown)


sveyret:
@hishamhm, I don’t think the failing test is caused by my modification. If it is, please tell me. If not, I’d be glad to have a review.

lenscas:
I think you can update https://github.com/teal-language/tl/blob/602943fa2ef4901c42f27679cbdd135a7066af5c/.github/workflows/playground.yml#L12 to point to node version 14 and have it work.

Seems like some node package got updated and it now requires node 14 instead of 12. Considering that this is error is while building a new playground version and more specifically the node.js parts rather than the lua/teal parts it is indeed not this change that broke things. I would also expect that any other PR will also fail with the same error.

spec/api/process_spec.lua
```diff
@@ -36,5 +36,25 @@ describe("tl.process", function()
          assert(foo_result, foo_err)
          assert(bar_result, bar_err)
       end)
+      it("should strip BOM from files", function()
```
hishamhm, line 39: Nice! Thank you for adding a test!

tl.lua
```diff
@@ -9850,6 +9850,15 @@ end
 
 
 
+local BOM = "\xEF\xBB\xBF"
+local function readFullFile(fd)
```
hishamhm, line 9854: minor coding style nitpick:

```suggestion
local function read_full_file(fd)
   local bom <const> = "\xEF\xBB\xBF"
```

Could you change it to use "snake_case" instead of "methodCase" for consistency with the rest of the file? Also, it should be fine to move the BOM constant inside the file and make it lowercase as well.

hishamhm (changes requested):
Thank you for the PR! Left some minor suggestions, I'll be happy to merge once those are addressed!

hishamhm:
@lenscas if you're able, could you push a PR that bumps the node version and confirms that this is the proper fix? otherwise I'll try to do it when I get the time. thanks!!

spec/api/process_spec.lua
sveyret: This is part of the job! :relaxed:

tl.lua
sveyret: Done

PR #574, opened 2022-10-13
Update playground.yml to use newer node.js version

lenscas:
As it goes wrong in the part that uploads the lua compiler to the playground I don't have a much better way to check if it works than making a PR and hope for the best.

I do however believe that updating the node.js version is the way forward compared to finding a version of netlify-cli that works on version 12 as 12 is ancient at this point.

### DIFF
--- a/.github/workflows/playground.yml
+++ b/.github/workflows/playground.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [12.x]
+        node-version: [14.x]
     steps:
     - uses: actions/checkout@v1
       with:


lenscas:
I guess the changed version doesn't run automatically, as it still tried to install node 12.

sveyret:
I made the same try and had the same problem… :disappointed:

Aire-One:
> I guess the changed version doesn't run automatically, as it still tried to install node 12.

Yeah. I think the used workflow is always the one in the repository default branch.

To try a new workflow, you have to fork the repository and change the fork default branch to the feature branch with the new workflow. So that GitHub picks it up on your forked repository.

hishamhm:
Thanks @lenscas! I'm moving forward and merging it

PR #575, opened 2022-10-15
playground: Use NodeJS 14.x as required by deps

sveyret:
Some dependencies now require usage of NodeJS 14.x


sveyret:
Sorry, I didn’t see you already made such a PR…

Issue #514, opened 2022-04-19
"not all enum values map to record fields of the same type" message proposal.


nagolove:
I have some code like this:
```lua
local enum Command
    "circle"
    "line"
    "rect"
end

local commands = {}

function commands.circle(_: string)
end

function commands.line(_: number)
end

function commands.rect(_: integer)
end

local cmd: Command = 'circle'
commands[cmd]()
```
And I get such message from compiler:
```
1 error:
example.tl:19:10: cannot index, not all enum values map to record fields of the same type
```
I have stumbled for time because of did not attended to function types. My idea is to write more complex message like:
```
example.tl:19:10: cannot index, not all enum values map to record fields of the same type( function(string) ~= function(number) )
```
In this message I have direct indication for first incompatible types pair.
Implementation proposal: apply text templates in `match_all_record_field_names` function argument like `"cannot index, not all enum values map to record fields of the same type( <f> ~= <t>)"` and replacing placeholders with real type names using`string.gsub()` .

hishamhm:
I suspect the errors could get pretty longwinded if they print the entire function signatures. Perhaps using their names, such as `(types of fields 'circle' and 'line' do not match)`?

nagolove:
@hishamhm Yes, only names are enough in most cases.

hishamhm:
Patches are welcome! :grin:

PR #576, opened 2022-10-16
Improve error message for inconsistent fields (#514)

sveyret:


spec/operator/index_spec.lua
```diff
@@ -50,7 +50,7 @@ describe("[]", function()
          local e: Keys = "foo"
          local n: number = x[e]
       ]], {
-         { msg = "cannot index, not all enum values map to record fields of the same type" },
+         { msg = "cannot index, not all enum values map to record fields of the same type (types of fields" },
```
hishamhm, line 53: is it possible to add the full message here with the field names? or is the check order non-deterministic? (I tried to catch all ordering issues a while ago; since the test runs over `ipairs` of the field names, they should return the same results every time)

hishamhm:
I've merged the playground fix from @lenscas so if you rebase and force-push your branch, the playground test should pass hopefully!

spec/operator/index_spec.lua
sveyret: Indeed, I was not sure the order was deterministic. OK, I will add the full message.
(for the moment, I just rebased to see if the check was working)

sveyret: Done…

hishamhm:
Sweet, thank you!

PR #585, opened 2022-11-23
Fix a case userdata record is not resolved properly

pigpigyyy:
The problem appeared after the new return inference commit, the minimal issue case is:
```lua
-- file: mod.tl
local record R
   userdata
end
local r: R
return r

-- file: foo.tl
local r = require("mod")
return r
```


hishamhm:
@pigpigyyy Thank you!!

PR #586, opened 2022-11-23
Fix returning record with type alias issue

pigpigyyy:
The problem appeared after recent commit. Minimal reproducing codes are:
```lua
-- file: mod.tl
local record R
   n: number
end
local record Mod
   type T = R
end
local inst: Mod
return inst

-- file: merged.tl
local mod = require("mod")
return {
   mod = mod
}

-- file: foo.tl
local merged = require("merged")
local t: merged.mod.T
print(t.n)
```
Tried fixing it, if there is better way of fixing, this PR can be just dropped.


hishamhm:
@pigpigyyy Merged manually, thank you!

PR #595, opened 2022-12-14
Add playground running in Lua compiled to wasm

mingodad:


mingodad:
This playground can be viewed here https://mingodad.github.io/tl

lenscas:
First 2 things I noticed: 

1: All examples seem to have 1 error reported however, I am not sure if those errors are correct?

2: There doesn't seem to be a way to share code with this editor.

3: Most examples don't seem to go much further than "this is how you declare a type", 
For the enum one for example, it might be nice also define a function that takes a parameter of this enum and just prints it, combined with calling this function directly, putting an enum variant in a variable and then calling said function and an commented out line that tries to call this function with a string that isn't part of the enum.

Other than that the editor does look good and looks to be a real improvement over the old one :)

mingodad:
The error indication is because it's using `ACE` `Lua` parser for syntax.

hishamhm:
@mingodad We do not keep the playground in the compiler repository. I think it is fine to keep it in your own repo, no need to merge it here — but in any case, thanks for showcasing another option for running Teal in the browser!

lenscas:
or, said another way. Perhaps a better place for this pr would've been https://github.com/teal-language/teal-playground

I do think it has the potential to be better than the current playground. However, the 3 things I noticed are big regressions that I think would need to be addressed before replacing the current one with this one can be done.

PR #604, opened 2022-12-30
Fix a case union with typearg in return value is not resolved properly

pigpigyyy:
Recent Teal commits are breaking an small case in my project. Took some time debugging around to find that when `check_args_rets` passes most of the args and returns checking, the `mark_invalid_typeargs` function will still mark typearg in union to be unresolvable.

### DIFF
--- a/spec/call/generic_function_spec.lua
+++ b/spec/call/generic_function_spec.lua
@@ -523,5 +523,17 @@ describe("generic function", function()
       local _foo: integer = Container:resolve()
       local _bar: string = Container:resolve()
    ]])
+
+   it("should resolve union with typearg in return value", util.check [[
+      local record A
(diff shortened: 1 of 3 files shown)


hishamhm:
@pigpigyyy Thank you for the PR with the testcase and initial solution! I took a _long_ time investigating this. I'm not 100% sure about my solution either, but I think it is a little more complete. You can compare the difference by running `TL_DEBUG=1 ./tl check 604.tl` and seeing the types resolved by the intermediate nodes (in particular, with a930e41, the "op @funcall" node resolves to `(A | nil)`, instead of resolving to `(T | nil)` and being ignored later.

As always, let me know if this causes any other problems! Thank you again!!

PR #627, opened 2023-02-05
Add optional third arg for string.rep

### LINKED ISSUE #593: Missing separator argument for string.rep
https://www.lua.org/manual/5.4/manual.html#pdf-string.rep
image

as you can tell i don't know how to write issues

icy-lava:
Fixes #593

Originally I thought I had to trigger the 5.3 compat in the definition of the function (since third arg is not present in Lua 5.1), but upon further inspection of the code it seems like it already gets triggered for any function in the `string` namespace.


hishamhm:
@icy-lava Merged! Kudos on your first Teal PR! :tada:

PR #642, opened 2023-03-20
Feature/record body declared methods

JR-Mitchell:
### What?
- Adds new `Type.is_self` property for identifying a function/method parameter that is `self`
- Adds new `string` parameter `name` for passing the name of the body / type being parsed to:
    - the `ParseBody` type
    - `parse_newtype`
    - `parse_record_body`
- Passes the record name using the `name` parameter for all cases where `parse_record_body` is called
- Updates `parse_argument_type` so that if the argument is named `self`, the type is marked as `is_self = true`
- Updates `parse_function_type` so that if the first argument has `is_self == true`, the function type is marked as `is_method = true`
- Updates `parse_record_body` so that, before a method is added to the record's fields, it is verified that the first argument matches the type of the record, and if it does not then it is marked as `is_self = false` and the function type as `is_method = false`
- Updates `recurse_type` so that the first argument is also recursed in a method if that argument is marked as `is_self == true`
- Consolidates the near-indentical `shallow_copy` and `shallow_copy_type` methods into one

### Why?
See #638 - this PR allows for forward declaration of methods in records, based on the first argument being named `self` and being correctly typed.
This allows for warnings and clearer errors when using types defined in `.d.tl` files, as well as allowing methods to be declared in one scope and then defined in another without losing these warnings.

### Anything else?
There is an edge case which may need discussion: the code
```
local record MyRecord<T>
    add: function(self: MyRecord<integer>, other: MyRecord<integer>)
end
local first: MyRecord<integer> = {}
local second: MyRecord<integer> = {}
first.add(second)
```
does not raise a warning, since `add` is not treated as a method for `MyRecord`, even though it could be argued that it should be treated as a method of the type realisation `MyRecord<integer>`.
I have a vague idea of how this could be achieved if we want this behaviour, but haven't been able to properly implement it yet.
Let me know whether it is fine that there is no warning in this situation, whether this can go into a separate issue / PR, or whether I should try to resolve it in this PR.

### DIFF
--- a/spec/call/record_method_spec.lua
+++ b/spec/call/record_method_spec.lua
@@ -64,35 +64,101 @@ describe("record method call", function()
       }))
    end)
 
-   it("catches wrong use of self. in call", util.check_type_error([[
-      local record Foo
-      end
-      function Foo:method_a()
-      end
-      function Foo:method_c(arg: string)
-      end
-      function Foo:method_b()
-         self.method_a()
-         self.method_c("hello")
-      end
-   ]], {
-      { y = 8, msg = "invoked method as a regular function: use ':' instead of '.'" },
-      { y = 9, msg = "invoked method as a regular function: use ':' instead of '.'" },
-   }))
+   describe("catches wrong use of self. in call", function()
+      it("for methods declared outside of the record", util.check_type_error([[
+         local record Foo
+         end
+         function Foo:method_a()
+         end
+         function Foo:method_c(arg: string)
+         end
+         function Foo:method_b()
+            self.method_a()
+            self.method_c("hello")
+         end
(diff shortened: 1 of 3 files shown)


hishamhm:
> There is an edge case which may need discussion

I think this is fine. The best we could do in that case would be to add a warning hint if the type (including type arguments) for a first argument called `self` is not nominally the same as the record the function is in, but for such specific scenario the programmer might want to call that `self` consciously and we'd be drifting into coding style, so better not make the type checker that opinionated.

Overall the PR looks good as far as I can tell! Merging, thanks!!

hishamhm:
also: released in Teal 0.15.2!

PR #649, opened 2023-04-13
Add new warning when nil is assigned to a non-nilable type

Hazematman:
This pull request add a new warning when you assign `nil` to a type that is not-nilable. For example the following code

```
local x: number = 10
x = nil
```

will now return an warning that looks like

```
1 warning:
test.tl:18:1: assigning nil to a non-nilable type number
```

hinting that the correct way to write this code is (assuming that you want assigning nil to be valid)
```
local x: number | nil = 10
x = nil
```

This change causes the tl compiler to throw a lot of warning when building. If required I could modify the tl compiler to explicitly mark types as nilable where required.

This change also required a test update to the record function spec tests, as one of the tests would assign nil to a boolean value and then expect zero warning to be generated.

### DIFF
--- a/spec/declaration/record_function_spec.lua
+++ b/spec/declaration/record_function_spec.lua
@@ -101,7 +101,7 @@ describe("record function", function()
            func: function(R)
          end
 
-         function R:func(): boolean
+         function R:func(): boolean | nil
            return self._current and self._current:func()
          end
       ]], {}, {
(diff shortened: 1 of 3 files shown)


Hazematman:
Closing this PR as it only address one issue surrounding nil ability of types. Proper nil checking would require more changes to the type system.

PR #668, opened 2023-06-09
Fix/remove pcall method warning

### LINKED ISSUE #656: Warning when calling method via pcall
When calling a method via pcall, the Teal compiler produces a warning that the method is being invoked as a regular function.

Example:
```
local record Test
    someString: string
end

function Test.new(someString: string): Test
    return setmetatable({ someString = someString }, { __index = Test })
end

function Test:print(): nil
    if self.someString == nil then
        error("someString is nil")
    end

    print(self.someString)
end

local test = Test.new()
local ok, err = pcall(test.print, test)
if not ok then
    print(err)
end
```

Output:
```
1 warning:
test.tl:18:22: invoked method as a regular function: consider using ':' instead of '.'
```

JR-Mitchell:
### What?
- Adds tests for the incorrect warnings and errors raised in #656 
- Updates `special_pcall_xpcall` so that the function type of the first argument is treated as not a method, even if it is a method

### Why?
Resolves #656 

`pcall` (and `xpcall`) invokes its first argument as a regular function with the provided arguments - it has no way of implicitly passing the `self` parameter from the method owner of its first argument to the method.
Therefore, it should not warn `invoked method as a regular function: use ':' instead of '.'` when its first argument is a method, or raise this as the error message when passed an argument of incorrect type.

### Anything else?
I was slightly worried that setting `is_method` to `false` on this type might cause unintended side-effects, however I have checked through the sequence of function calls from `special_pcall_xpcall` where `ftype` is passed, and it appears that `is_method` is exclusively used for determining whether to show these warnings / errors.

### DIFF
--- a/spec/stdlib/pcall_spec.lua
+++ b/spec/stdlib/pcall_spec.lua
@@ -72,4 +72,21 @@ describe("pcall", function()
    ]], {
       { msg = "xyz: got boolean, expected number" }
    }))
+
+   it("does not warn when passed a method as the first argument", util.check_warnings([[
+      local record Text
+         text: string
+      end
+      function Text:print(): nil
+         if self.text == nil then
+            error("Text is nil")
+         end
+         print(self.text)
+      end
+      local myText: Text = {}
+      local ok, err = pcall(myText.print, myText)
+      if not ok then
+         print(err)
+      end
+   ]], {}, {}))
 end)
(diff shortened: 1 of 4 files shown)


hishamhm:
@JR-Mitchell nice! I'm glad the solution turned out not convoluted!

> I was slightly worried that setting is_method to false on this type might cause unintended side-effects, however I have checked through the sequence of function calls from special_pcall_xpcall where ftype is passed, and it appears that is_method is exclusively used for determining whether to show these warnings / errors.

I feel you — that whole `argdelta` logic for shifting arguments around for methods vs non-methods vs pcall gets real tricky and I struggle to follow it myself when it's been too long since I last looked at it, but the test cases give us confidence that the behaviors are good.

PR #669, opened 2023-06-09
record functions inherit type variables from the record definition

### LINKED ISSUE #657: Cannot use generics with methods
It's not possible to use methods with generic records:

Code:
```
local record Test<T>
    value: T
end

function Test.new<T>(value: T): Test<T>
    return setmetatable({ value = value }, { __index = Test })
end

function Test:print()
    print(self.value)
end
```

Output:
```
2 errors:
test.tl:9:10: missing type arguments in record<T> (value: T; new: function<T>(T): Test<T>; print: function(self))
test.tl:10:16: cannot index key 'value' in bad_nominal 'self' of type Test
```

The compiler should automatically assume that self is of type `Test<T>`.

hishamhm:
For records that have type variables (e.g. `local Record Test<T>`), implictly declare those type variables (`T`) in the scope of record functions and methods (e.g. you can use `T` in the body and arguments of `function Test:my_method() --[[...]] end`)

Fixes #657.

### DIFF
--- a/spec/declaration/record_method_spec.lua
+++ b/spec/declaration/record_method_spec.lua
@@ -460,6 +460,21 @@ describe("record method", function()
       end
    ]]))
 
+   it("inherits type variables from the record definition (regression test for #657)", util.check([[
+      local record Test<T>
+          value: T
+      end
+
+      function Test.new(value: T): Test<T>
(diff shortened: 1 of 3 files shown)


Frityet:
works great for me! 👍

PR #674, opened 2023-07-08
fix string.find signature

fperrad:
return only 2 integers as start and end position


hishamhm:
>  If the pattern has captures, then in a successful match the captured values are also returned, after the two indices.

That is why the return is of vararg type `integer, integer, string...`

hishamhm:
@fperrad no worries! thanks for giving Teal a spin!!

PR #686, opened 2023-08-01
generate bit32 with Lua 5.2, and bit with Lua 5.1 & LuaJIT

fperrad:
My point is that's a pity to doesn't generate the best code for LuaJIT.

This introduces a new value for the option `--gen-target`.

### DIFF
--- a/docs/compiler_options.md
+++ b/docs/compiler_options.md
@@ -23,7 +23,7 @@ return {
 | `-l --require`  | | `{string}` | `run` | Require a module prior to executing the script. This is similar in behavior to the `-l` flag in the Lua interpreter. |
 | `-I --include-dir` |  `include_dir` | `{string}` | `build` `check` `gen` `run` | Prepend this directory to the module search path.
 | `--gen-compat` | `gen_compat` | `string` | `build` `gen` `run` | Generate compatibility code for targeting different Lua VM versions. See [below](#generated-code) for details.
-| `--gen-target` | `gen_target` | `string` | `build` `gen` `run` | Minimum targeted Lua version for generated code. Options are `5.1`, `5.3` and `5.4`. See [below](#generated-code) for details.
+| `--gen-target` | `gen_target` | `string` | `build` `gen` `run` | Minimum targeted Lua version for generated code. Options are `5.1`, `5.2`, `5.3` and `5.4`. See [below](#generated-code) for details.
 || `include` | `{string}` | `build` | The set of files to compile/check. See below for details on patterns.
 || `exclude` | `{string}` | `build` | The set of files to exclude. See below for details on patterns.
 | `-s --source-dir` | `source_dir` | `string` | `build` | Set the directory to be searched for files. `build` will compile every .tl file in every subdirectory by default.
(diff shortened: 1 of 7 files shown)


hishamhm:
@fperrad One concern I had originally was that there were some semantics differences between `bit` and `bit32` (I don't remember the details, but I remember some significant arguments between Mike and Roberto in lua-l back in the day), so my rationale was that sticking with `bit32` meant that the Teal operators would have the most similar semantics regardless of targetted VM (modulo inevitable differences between 32 and 64 bit numbers). Any thoughts on this point?...

alerque:
I'm not 100% sure that there are no semantic differences now, but am unaware of anything but syntax differences. I've been using [this bitshim module](https://github.com/sile-typesetter/sile/blob/master/lua-libraries/bitshim.lua) to provide a normalized syntax that can be driven by either bit, bit32, or the new operators in Lua 5.3+.

hishamhm:
@alerque There are. This thread contains at least one example and more info (with posts from Roberto and Mike Pall) http://lua-users.org/lists/lua-l/2010-10/msg00286.html

The difference described by David Manura in that post still applies:

```
] luajit
LuaJIT 2.1.0-20220411 -- Copyright (C) 2005-2022 Mike Pall. https://luajit.org/
JIT: ON SSE3 SSE4.1 BMI2 fold cse dce fwd dse narrow loop abc sink fuse
> print(bit.lshift(0xffffffff, 32))
-1

] lua
Lua 5.3.6  Copyright (C) 1994-2020 Lua.org, PUC-Rio
> bit32.lshift(0xffffffff, 32)
0
```

hishamhm:
@fperrad what Teal currently does _is_ my best effort. ;)

It is not trying to be magical. Perhaps the difference is that you may be looking at Teal as a Lua-code-generating-tool. I look at it as a programming language. As a language, it has its own semantics, which are my responsibility to define (or to leave undefined — or in the words of the great Neil Peart, "if you choose not to decide, you still have made a choice").

Authors of Teal libraries will be writing Teal code and getting one given semantics for a given number size. It's fair to assume that Teal module authors will expect that consumers of their libraries will get the author-intended semantics wherever they run their code.

On a more personal note, one of the things that do annoy me in the Lua ecosystem are the myriad small inconsistencies across versions. I've spent many years writing LuaRocks code that keeps walking on eggshells to be able to run on every relevant Lua version; I'd rather not have to — I want to have one semantics I know, code for that, and have it "just work". This was probably the secondary motivation for Teal (the first one being type-checking). That's also why it loads the compat libraries by default.

I am still reluctant about adding a mode that makes the bitwise operators generate lua-bitop, but if we do add that, it would have to be a flag of its own, and definitely not the default target mode for Lua 5.1. After all, IIRC lua-bitop can be compiled and used with other Lua versions too. I also don't want to associate it to `--gen-target=luajit`, because the generation target shouldn't alter the language semantics (modulo unavoidable changes such as integer size).

I hope you understand these concerns of mine are coming from a place of responsibility regarding the stewardship of the language, rather than any preference between bit32 and lua-bitop!

PR #687, opened 2023-08-03
fix file:close signature

fperrad:
since Lua 5.2, when closing a file handle created with `io.popen`, `file:close` returns the same values returned by `os.execute`.

### DIFF
--- a/tl.lua
+++ b/tl.lua
@@ -5227,7 +5227,7 @@ local function init_globals(lax)
             typename = "record",
             is_userdata = true,
             fields = {
-               ["close"] = a_type({ typename = "function", args = TUPLE({ NOMINAL_FILE }), rets = TUPLE({ BOOLEAN, STRING }) }),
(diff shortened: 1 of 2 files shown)


hishamhm:
It would be good to get this merged and released when merging this here!...
https://github.com/lunarmodules/lua-compat-5.3/pull/53

I just sent a ping over there. If I get no response I'll go ahead and merge both.

hishamhm:
I got no response regarding compat53 from @siffiejoe, so I went ahead and moved the [LuaRocks entry](https://luarocks.org/modules/lunarmodules/compat53) to the [lunarmodules](https://luarocks.org/modules/lunarmodules) org (of which [we're both members on GitHub](https://github.com/orgs/lunarmodules/teams/team-lua-compat)), and then [released the update](https://luarocks.org/modules/lunarmodules/compat53/0.11-1) which includes the change needed by this PR.

PR #702, opened 2023-09-27
fix io.write & file:write signature

fperrad:
quoted from https://www.lua.org/manual/5.4/manual.html#pdf-file:write
> Writes the value of each of its arguments to `file`. The arguments must be strings or numbers.


hishamhm:
Thank you! TIL!

Issue #713, opened 2023-10-30
debug.getlocal -- fixes to signature and return values


JLPLabs:
The code below, using `debug.getlocal([thread,] f, local)`, generates warnings and errors, despite using `getlocal` as specified in [1] (and all previous versions through and including 5.1). Quoting from [1], _This function returns the name and the value of the local variable with index local of the function at level f of the stack._

```
-- getlocal.tl
local name: string
local val: string
name, val = debug.getlocal(2, 1)
```
`tl check` generates...
```
tmp> tl check getlocal.tl 
========================================
2 warnings:
getlocal.tl:3:1: only 0 values are returned by the function
getlocal.tl:3:7: only 0 values are returned by the function
========================================
3 errors:
getlocal.tl:3:1: variable is not being assigned a value
getlocal.tl:3:7: variable is not being assigned a value
getlocal.tl:3:28: argument 1: got integer, expected function(...: <any type>): <any type>...
```

getlocal is defined at line 5313 in tl.tl.
```
            ["getlocal"] = a_type {
               typename = "poly",
               types = {
                  a_type { typename = "function", args = TUPLE { THREAD, FUNCTION, NUMBER }, rets = TUPLE {} },
                  a_type { typename = "function", args = TUPLE { FUNCTION, NUMBER }, rets = TUPLE {} },
               },
            },
```

Perhaps it is appropriate to insert another definition to types{} table, like this:
```
                  a_type { typename = "function", args = TUPLE { NUMBER, NUMBER }, rets = TUPLE { STRING, STRING } },
```

I'd be happy to do a pull request if that is your preference.

---

[1] Lua 5.4 Reference Manual; [https://www.lua.org/manual/5.4/manual.html#pdf-debug.getlocal](https://www.lua.org/manual/5.4/manual.html#pdf-debug.getlocal)

hishamhm:
A PR is definitely welcome!

hishamhm:
it's weird that the other cases of the definition are also missing
`rets`... Also, in the case you're adding, the rets tuple should probably
be something like `STRING, ANY`?

(Replying via email from the phone, so sorry for any bad formatting!)

JLPLabs:
Will do.
Good point about STRING, ANY.

JLPLabs:
to your point about the other cases needing 'rets'... I could do a PR for them such that the entire branch looks like this

```
types = {
[...]

PR #714, opened 2023-10-30
fix signature and return values for debug.getlocal; issue #713

JLPLabs:
I tested the fix with my use case. (I did indeed take your suggestion to return STRING, ANY.
I did not do any further tests.


JLPLabs:
my commit from today (2023.11.01 -- which fixes 'rets' for all debug.getlocal() cases, seems to have failed the playground test. Not sure what the error means.

hishamhm:
I regenerated `tl.lua` to add all cases and merged manually, thank you!

PR #718, opened 2023-11-05
fix io.read & file:read signature

fperrad:


hishamhm:
I think it's better to declare this as a polymorphic function so we don't have to use casts in the most common cases. I will try my hand at an alternative implementation, but thank you for the PR pointing out the issue!

PR #730, opened 2024-01-01
hashbang, if exists and if wanted, stays on line 1

JLPLabs:
fixes 729 and adds the proposed test

### DIFF
--- a/spec/cli/gen_spec.lua
+++ b/spec/cli/gen_spec.lua
@@ -1,6 +1,7 @@
 local util = require("spec.util")
 
 local input_file = [[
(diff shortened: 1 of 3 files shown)


hishamhm:
Thanks for the PR! That Windows CI error looks unrelated, but I'll still try to take a look at that prior to merging -- I'm on my phone right now but I'll look at this next time I'm at the computer.

hishamhm:
I noticed another test case which indicated a need for a further adjustment, so I based the extra changes off your test case and merged it directly! Thanks!

PR #711, opened 2023-10-21
experiment: make type system more nominal

hishamhm:
Instead of treating nominal records nominally and all other nominal types structurally, with this commit we treat all nominal types nominally except for unions, which are treated structurally.

Give this branch a try in your codebase and let me know your impressions!

[Marked as "draft" because I do not intend to merge this as-is, as there are no test cases for the different behaviors, etc.]

Issue #708, opened 2023-10-20
Potential issue with two types using the same primitives


FourierTransformer:
I'm trying to use Teal to have the concept of units, and while I can create my own new types based off of a `number`, however I'm allowed to still add those two different types together.

Here's an example:

```teal
local type Temperature = number
local type Day = number

local temp: Temperature = 45
local birthday: Day = 34

print(temp + birthday) -- doesn't report as an error or warning
```

I'm not exactly sure if this is a bug, but I feel like you should only be able to get away with this if you use `as` to have it be treated as a different type.

FourierTransformer:
Ah, I see how this conflicts with what `is` does, and overall could cause a small can of worms.

```
local result:Day = temp + birthday
print(result is Temperature) -- which ends up being true since type(result) == "number"
```

hishamhm:
@FourierTransformer yeah, this use case is not currently covered by the type system.

The underlying concept you're looking for is whether the type system is _nominal_ or _structural_. In a nominal type system, types with different names are different types. In a structural type system, types with the same data shape are the same type.

Teal's type system, as those from most programming languages, is a mix of both. For instance, record types are nominal, but function types and other primitive data types are structural.

There are languages that make a distinction between "new type declarations" and "type alias declarations", where the former declares a new distinct incompatible type and the latter declares a new name for the same type. (Haskell is especially confusing in this regard because `type` means a type alias, `data` means a new type, and `newtype` is a [special-case new type which is erased at runtime](https://wiki.haskell.org/Newtype) so behaves more like an alias).

So, to keep things simple for the common cases, `type` in Teal means a new type for nominal types such as records, and a type alias for structural types:

So defining a `type` as a shorthand alias for a function works as expected:

```lua
local type F = function(number): string

local function uses_f(fn: F)
   print(fn(10) .. "!")
end

local my_f = function(n: number): string
   return "hell" .. ("o"):rep(n)
end

-- my_f was not declared as an F, but it's of the correct shape
uses_f(my_f) -- prints "helloooooooooo!"
```

Functions being structurally checked, this also works:

```lua
local type F = function(number): string

local function uses_f(fn: F)
   print(fn(10) .. "!")
end

local type G = function(number): string

local my_g = function(n: number): string
   return "hell" .. ("o"):rep(n)
end

-- my_g is a G, not an F, but they're the same shape
uses_f(my_g) -- prints "helloooooooooo!"
```

On the other hand, records are nominal, so this is not allowed:

```lua
[...]

PR #732, opened 2024-01-08
Next version!

hishamhm:
Checklist for next version release:

* [x] usability milestones
   * [x] LuaRocks codebase converted to Teal
   * [x] Teal codebase itself uses interfaces
* [x] ensure that everyone who's on the current version has a clear path to migration
   * [x] ensure APIs from `master` are compatible (vscode-teal, Cyan, Defold, etc.)
      * [x] +1 from vscode-teal
      * [x] +1 from Cyan
      * [x] +1 from teal-language-server
   * [x] pragma integration to avoid source-level compatibility issues (feat-arity)
   * [x] ideally no known regressions (no open bugs that are `next`-only)


bjornbm:
I'm excited for this new release, but struggle to grasp the `interface`s and probably other changes. Basically I don't see much updates/explanation in `tutorial.md`, neither of interfaces not the move to nominal types (#711, #708).

euclidianAce:
From a quick glance I like the api changes and interfaces are a huge boon. Writing up some quick and dirty [luv](https://github.com/luvit/luv) definitions (which is full of inheritance/interfaces) seems to be working well. I see `interface` `where` clauses are implemented as macros, which are also exposed via a new `local macroexp foo(...)` statement which after giving some thought is probably the best/lowest-amount-of-tying-users-of-teal-to-one-object-model way for a language like teal to add that sort of feature. Also nominal types are always a :+1: from me.

I do need to get back into the swing of working on teal related stuff, but I definitely fell into the got-a-real-job-and-now-have-less-time-to-work-on-hobby-projects trap that happens so often with open source.

All in all, looks great to me!

catwell:
Same as @euclidianAce, real life (work + new kid)  means I have little time for OSS but the changes I see in this branch are a lot of things I thought would be great for Teal.

Once it is a bit more stable I'll try it on my existing code bases in Teal.

hishamhm:
@bjornbm tests and documentation are still pretty much missing, because I wanted to iterate as quickly as possible to experiment with the usability of these features... but once the dust settles we definitely need them!

hishamhm:
@euclidianAce @catwell Thanks a lot for the feedback!! And, yeah, I totally get that sentiment — over the years I've accepted that my work on FOSS projects happens in bursts. When I find that I get a window of availability and motivation to work on this stuff, I try to get the most out of it. :)

For Teal, having a quiet time for the project since the last burst was a good thing as it allowed me to set priorities: seeing what are the gaps that people most often try to solve, etc. I think it's clear by now that the top two things missing in current Teal are interfaces and better `nil` handling. I don't think we'll get non-nilable types just yet, but it's the next thing on my mind. With [`--feat` flags ](https://github.com/teal-language/tl/pull/732/commits/44b64665d48370249cac65a12e6a2914eb66a675) it should be easier to make gradual changes to the language in the future. [This writeup](https://github.com/seanbaxter/circle/blob/master/new-circle/README.md) has made the rounds recently and makes a good case for gradual language evolution via per-compilation-unit feature flags.

hishamhm:
Also, speaking of writeups, it's funny that CI is failing because the `tl.type_check` function is exceeding the maximum number of upvalues per function in Lua 5.1. Looks like old Lua wasn't a fan of [Carmack's long function style](http://number-none.com/blow/john_carmack_on_inlined_code.html); I'll have to do some refactoring!

hishamhm:
Update: it took me a big reorganization in [d885f90](https://github.com/teal-language/tl/pull/732/commits/d885f90a0a88d21713c9f2b1b3d838a850ac6599) to get CI green. This is because the size of tl.tl started hitting limits in both Lua 5.1 (number of upvalues) and Lua 5.4 (number of locals). The goal of the changes in this latest commit is to be able to finally break tl.tl into multiple files, but I wanted to do the reorg first to produce a sensible diff (which would let me more easily track the changes). The commit description includes more details on the internal changes.

hishamhm:
Regarding "making the type system more nominal", here's one interesting example:

https://toot.kottman.xyz/@michal/111860362351389256
> One thing I'm missing (or haven't yet found in docs) is "distinct type" (borrowing term from Nim) or integer based enum - e.g. `local type LibraryErrorCode = distinct integer; function work():LibraryErrorCode`

https://mastodon.social/@hisham_hm/111861339267098398
> if you create a new nominal type for a base type like integer, it will be distinct across nominal types, but still be compatible with plain integer (so you can use the math. library etc)

This checks without errors in the `master` branch, but will check for `Code` against `Age` here in `next`:

```lua
local type Code = integer

local function get_code(): Code
end

local type Age = integer

local function get_age(): Age
end

local x: integer = get_code()
x = get_age()

local c: Code = get_code()
c = get_age()                         -- ERROR: in assignment: Age is not a Code

local a: Age = get_age()
a = get_code()                        -- ERROR: in assignment: Code is not a Age
```

catwell:
I played a bit with the `where` clause on the current `next` branch, and I noticed that it doesn't work with generics. I mean that this works:

```
local type Success = record
    where self.error == false
    error: boolean
    value: integer
end
```

but not this:

```
local type Success = record<T>
    where self.error == false
    error: boolean
    value: T
end
```

It results in:

```
example.tl:2:11: missing type arguments in record<T>
example.tl:2:16: cannot index key 'error' in invalid 'self' of type self
```

Is that intended or do you intend to support it?

hishamhm:
@catwell thanks for the test case! I haven't really worked on the relationship between generics and interfaces/`where` yet, so that's a good starting point.

> Is that expected or do you intend to support it?

I intended `where` to work alongside `is`, as in `record __ is __ where`, but having `where` on its own might be workable for unions. As for the generic there, I intend to support this test case, but I expect I'll need to place some limitations in the interactions between `is ___` subtyping and generics. I'll probably over-restrict to be on the safe side first, then see what can be relaxed.

catwell:
> (Slightly related: can you release your FOSDEM slides? I know the video is probably broken, which is too bad, but the slides at least would be useful.)

The video [was *not* broken](https://fosdem.org/2024/schedule/event/fosdem-2024-2564-five-years-of-teal-minimalism-versus-growth-in-language-design/) \o/

uramer:
How are `where` and `__is` intended to work for type declarations?

docs/tutorial.md
```diff
@@ -180,9 +179,8 @@ local record File
    close: function(File): boolean, string
 end
 
--- an arrayrecord: a record which doubles as a record and an array
-local record TreeNode<T>
-   {TreeNode<T>}
+-- a record can doubles as a record and an array, by declaring an array interface
```
darrenjennings, line 182:
```suggestion
-- a record can double as a record and an array, by declaring an array interface
```

hishamhm:
I think we're in a good shape to merge this into `master`!

* I've gotten positive feedback from @euclidianAce in https://github.com/teal-language/cyan/pull/46 — I guess that counts as a +1?
* @pdesaulniers vscode-teal support looks good for `next` as well, right?
* @svermeulen could you check if this works with teal-language-server? (existing Teal code might need `--feat-arity=off` to work out of the box but otherwise it should!)

Last thing before the new release will be merging this, writing some docs about interfaces,then we're good to go!

svermeulen:
@hishamhm will do.  I'll also start migrating my code base to `next` branch and report any issues I find there

euclidianAce:
I've just pushed my `tl-next` branch of cyan that I've been working with locally and things are working great. So this is a :+1: from me

svermeulen:
I can confirm tls working well with tl next with just some minor fixes.

Would it be possible to get preview version of tl published to luarocks?  Then I can do the same with tls. Or alternatively I can just wait until it merges and use an official release.

Btw - Really appreciate the arity and interface features. After upgrading it's clear these will be very useful in our codebase.  Upgrading was pretty straightforward for us - even caught some potential runtime bugs after enabling the arity check

hishamhm:
> Would it be possible to get preview version of tl published to luarocks? 

You can do `luarocks install tl --dev --branch next`, is that sufficient?

> Or alternatively I can just wait until it merges and use an official release.

I think that should be fine. It should be a matter of days!

> Btw - Really appreciate the arity and interface features. After upgrading it's clear these will be very useful in our codebase. Upgrading was pretty straightforward for us - even caught some potential runtime bugs after enabling the arity check

That is great to hear! Thank you so much for the positive feedback, it means a lot!!

hishamhm:
@pdesaulniers I'm assuming it's a +1 from vscode-teal as well, since you've added the extra syntax highlighting in 0.10.0, and @V1K1NGbg has used vscode-teal 0.10.0 happily with tl `next` in his LuaRocks porting project for quite a while!

pdesaulniers:
@hishamhm Autocompletion behaves a bit strangely right now, but I haven't had the time to debug this further... (Also, I'm not sure if these issues are actually new to the `next` branch!)

I can at least reproduce this specific case:

```
-- test.tl

local function hello(): number
    return 1
end

if 1 == 1 then
    local abc = hello()
    local def = abc
    def = abc
    




end
```

```
$ ./tl types -p 12:1 test.tl
{"hello":9}
```

 It seems like the two local variables `abc` and `def` are missing in the output. I'm running `tl` at commit [58ebd4b](https://github.com/teal-language/tl/pull/732/commits/58ebd4b7ab2c57dcd19f5c5e29c69221d8a016aa).

catwell:
> now the whole adventofcode repo typechecks on `next` with `--feat-arity off`!

2021 does, there are still a few errors on 2020 :)

In particular it does not like when I call this with a single argument (e.g in day13 part1):

```
function M.parse_integers(s: string, i0: integer) : {integer}
    local t, p = {}, i0 or 1
    local function f(x: string)
        t[p] = math.tointeger(x)
        p = p + 1
    end
    s:gsub("[-%d]+", f)
    return t
end
```

In day20 there is a different error probably related to the "empty table" issue:

```
local function count_sea_monsters(image: {string}): integer, integer
    local c, m, n = 0, {{}, {}}, 0
    for row = 1, #image - 2 do
        m[row + 2] = {}
        for col = 1, #image[1] - 19 do
            if has_sea_monster(image, row, col) then
                c = c + 1
                mark_sea_monster(m, row, col)
            end
        end
    end
    if c then
        for row = 1, #image do
            for col = 1, #image[1] do
                if image[row]:sub(col, col) == "#" and (not m[row][col]) then
                    n = n + 1
                end
            end
        end
    end
    return c, n
end
```

```
./common.tl:238:22: in assignment: got {}, expected {} | {}
./common.tl:249:68: cannot index object of type {} | {} with integer
```

Line 238 is `m[row + 2] = {}`, line 249 is `if image[row]:sub(col, col) == "#" and (not m[row][col]) then`.

hishamhm:
> 2021 does, there are still a few errors on 2020 :)

I guess my `for`-loop in the shell missed a few entries then! Thanks for the test cases, I will take a look later.

hishamhm:
@pdesaulniers Interesting. This works:

```
-- test.tl

local function hello(): number
    return 1
end

if 1 == 1 then
    local abc = hello()
    local def = abc
    




    def = abc
end
```

...so it looks like it is ending the scope "too early"? Might be a simple fix. BTW, looks like it also happens on `master`, so I probably won't treat it as a blocker!

hishamhm:
@catwell 

```
for f in `find . -path './day*/p*tl'`; do ( cd $(dirname "$f"); pwd; echo $f; TL_BIN_PATH=$HOME/github.com/teal-language/tl/tl ../tl.sh run $(basename $f); echo ); done
```

:grimacing: 

All of 2020 and 2021 should be passing with `--feat-arity off` now!

hishamhm:
@pdesaulniers That test case should be now fixed! It seemed to be specific to if-blocks

pdesaulniers:
@hishamhm I think I found a `tl types` issue that is specific to the `next` branch. It seems like record 'methods' are missing in the output.

point.tl:

```lua
local record Point
    x: number
    y: number
end

function Point:init(x: number, y: number)
    self.x = x
    self.y = y
end
```

```
$ tl types point.tl > point.json
```

Before (master branch):

```json
        "4": {
            "fields": {
                "init": 2,
                "x": 1,
                "y": 1
            },
            "file": "point.tl",
            "str": "type Point",
            "t": 131080,
            "x": 1,
            "y": 1
        },
        "5": {
            "fields": {
                "init": 2,
                "x": 1,
                "y": 1
            },
            "file": "point.tl",
            "str": "Point",
            "t": 131080,
            "x": 1,
            "y": 1
        },
```

After (next branch):

```json
        "9": {
            "fields": {
                "x": 5,
                "y": 5
            },
            "file": "point.tl",
            "str": "record",
            "t": 131080,
            "x": 1,
            "y": 1
        },
        "10": {
            "fields": {
                "x": 5,
                "y": 5
            },
            "file": "point.tl",
            "str": "type record",
            "t": 131080,
            "x": 1,
            "y": 1
        },
```

'init' is missing.

hishamhm:
@pdesaulniers it should be fixed now! the construction of the type report now happens as part of the type checking pass. Records with functions are (IIRC!) the only types that "grow" over time as one reads through a file, and their type definitions were not being updated in the report (because the typeinfo entries are memoized). 6117c8c should fix this, testcase included.

pdesaulniers:
Great, it seems OK now!

I think I noticed another issue. It seems like the 'ref' field is missing for record fields in some cases.

```lua
local record Operator
    operator: string
end

local record Node
    node1: Node
    operator: Operator
end

local function node_is_require_call(n: Node): string
    if n.operator and n.operator.operator == "." then
       return node_is_require_call(n.node1)
    end
end
```

```sh
$ tl types point.tl > point.json
```

Simplified output (next branch):

```json
    "by_pos": {
        "test.tl": {
            "11": {
                "9": 12,
                "24": 12,
            },
        }
    },
    "types": {
        "9": {
            "fields": {
                "operator": 7
            },
            "file": "test.tl",
            "str": "Operator",
            "t": 131080,
            "x": 1,
            "y": 1
        },
        "12": {
            "file": "test.tl",
            "str": "Operator",
            "t": 268435456,
            "x": 15,
            "y": 7
        }
    }
```

In the master branch, the second "Operator" type had a "ref" field, which lead to the record declaration:

```
        "5": {
            "file": "test.tl",
            "ref": 6,
            "str": "Operator",
            "t": 268435456,
            "x": 15,
            "y": 7
        },
        "6": {
            "fields": {
                "operator": 1
            },
            "file": "test.tl",
            "str": "type Operator",
            "t": 131080,
            "x": 1,
            "y": 1
        },
```

PR #750, opened 2024-06-13
Pragmas

hishamhm:


Frityet:
Maybe this shouldn't be an error? I see the value of some pre-processing tools having custom pragmas

hishamhm:
From a comment in https://github.com/teal-language/tl/issues/786

This might be useful to add as documentation at some point:

> Also, the pragma proposal is not intended to double as annotations (the kinds of things you usually see with @- syntax in other languages). Pragmas are intended as compiler directives, more akin to compiler flags (e.g. the #pragma use in C compilers). I understand these sit in a gray area, but I don't think there's a well-established definition in the literature.

> My practical goal for pragmas is to allow for handling compatibility issues when dealing with the language evolution. That is, in a Teal codebase with no legacy concerns, there should be no pragmas.

hishamhm:
Merged into `next`!

PR #762, opened 2024-07-10
Codegen Lua 5.1 friendly string literals

euclidianAce:
This gets the job done, but is arguably inelegant. `gsub`ing a large string literal a few times will probably have some performance implications and feels hackish. And they get run for every string literal whether it has escapes or not. I think a better solution might be to handle escapes at lex time and stash a list of byte offsets in the `Token` for easier translation at codegen time. So consider this a functional first draft/mvp.

### DIFF
--- a/tl.lua
+++ b/tl.lua
@@ -4305,7 +4305,7 @@ function tl.pretty_print_ast(ast, gen_target, mode)
 
 
 
-            if node.tk:sub(1, 1) == "[" or gen_target ~= "5.1" then
+            if node.tk:sub(1, 1) == "[" or gen_target ~= "5.1" or not node.tk:find("\\", 1, true) then
                return emit_exactly(node, children)
             end
 
--- a/tl.lua
+++ b/tl.lua
@@ -10792,7 +10792,7 @@ end
 
 
 local function read_full_file(fd)
-   local bom = "\xEF\xBB\xBF"
+   local bom = "\239\187\191"
    local content, err = fd:read("*a")
    if content:sub(1, bom:len()) == bom then
       content = content:sub(bom:len() + 1)
--- a/tl.lua
+++ b/tl.lua
@@ -4305,7 +4305,7 @@ function tl.pretty_print_ast(ast, gen_target, mode)
 
 
 
-            if node.tk:sub(1, 1) == "[" then
+            if node.tk:sub(1, 1) == "[" or gen_target ~= "5.1" then
                return emit_exactly(node, children)
             end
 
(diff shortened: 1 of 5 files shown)


euclidianAce:
Ha, funnily enough that 5.1 test failure is what this is trying to solve.

hishamhm:
Nice!

> gsubing a large string literal a few times will probably have some performance implications and feels hackish. And they get run for every string literal whether it has escapes or not. 

I've added a check here: https://github.com/teal-language/tl/tree/euclidianAce-string-codegen-check

> I think a better solution might be to handle escapes at lex time and stash a list of byte offsets in the Token for easier translation at codegen time. 

I've also tried my hand implementing your suggestion, based on your code: https://github.com/teal-language/tl/tree/euclidianAce-string-codegen-escapes

I did some light benchmarking with `hyperfine` and I couldn't see clear performance differences (the variation caused by CPU throttling in my laptop was greater than the difference across branches).

What do think about these variations?

euclidianAce:
If there isn't a blatant performance degradation then this is probably fine to merge (barring any `gen_compat` checks that should be added).

tl.lua
```diff
@@ -4296,6 +4297,74 @@ function tl.pretty_print_ast(ast, gen_target, mode)
...
       },
+      ["string"] = {
+         after = function(node, children)
+
+
+
+
+
+
+            if node.tk:sub(1, 1) == "[" or gen_target ~= "5.1" then
```
hishamhm, line 4308:
```suggestion
            if node.tk:sub(1, 1) == "[" or gen_target ~= "5.1" or not string.find(node.tk, "\\", 1, true) then
```

I would still add this bit just to stay on the safe side performance-wise

tl.tl
```diff
@@ -4296,6 +4297,74 @@ function tl.pretty_print_ast(ast: Node, gen_target: TargetMode, mode: boolean |
...
       },
+      ["string"] = {
+         after = function(node: Node, children: {Output}): Output
+            -- translate escape sequences not supported by Lua 5.1
+            -- in particular:
+            --  - \z : removes trailing whitespace
+            --  - \xXX : hex byte
+            --  - \u{} : unicode
+
+            if node.tk:sub(1, 1) == "[" or gen_target ~= "5.1" then
```
hishamhm, line 4308:
```suggestion
            if node.tk:sub(1, 1) == "[" or gen_target ~= "5.1" or not string.find(node.tk, "\\", 1, true) then
```

I would still add this bit just to stay on the safe side performance-wise

hishamhm:
Merged!

PR #798, opened 2024-08-31
docs: pragmas

hishamhm:


docs/pragmas.md
```diff
@@ -0,0 +1,134 @@
...
+If you're coming from an older version of Teal, it is possible that you will
+start getting lots of errors related to numbers of arguments, such as:
+
+```
+wrong number of arguments (given 2, expects 4)
+```
+
+This is because, up to Teal 0.15.x, the language was lenient on the _arity_ of
+function calls (the number of expressions passed as arguments in the call). It
+would just assumed that any missing arguments were intended to be `nil` on
```
Tieske, line 26:
```suggestion
would just assume that any missing arguments were intended to be `nil` on
```

docs/pragmas.md
```diff
@@ -0,0 +1,134 @@
...
+purpose. More often than not, this is not the case, and a missing argument is
+does not mean that the argument was option, but rather than the programmer
+forgot about it (this is common when adding new arguments during a code
```
Tieske, line 29:
```suggestion
purpose. More often than not, this is not the case, and a missing argument
does not mean that the argument was optional, but rather that the programmer
forgot about it (this is common when adding new arguments during a code
```

hishamhm:
@Tieske thank you for the proofreading!

docs/pragmas.md
```diff
@@ -0,0 +1,134 @@
...
+purpose. More often than not, this is not the case, and a missing argument
+does not mean that the argument was optional, but rather that the programmer
+forgot about it (this is common when adding new arguments during a code
+refactor).
+
+Teal now features _optional function arguments_. if an argument can be
+optionally elided, you now can, or rather, have to, annotate it explicitly
+adding a `?` to its name:
+
+```
```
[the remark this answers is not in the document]
hishamhm: sure! good call

PR #815, opened 2024-10-12
fix math.frexp & math.ldexp signature

fperrad:
see https://www.lua.org/manual/5.2/manual.html#pdf-math.frexp
  & https://www.lua.org/manual/5.2/manual.html#pdf-math.ldexp


hishamhm:
@fperrad Thank you!!

Issue #824, opened 2024-10-15
Teal crashes when type checking this specific file


svermeulen:
Using the current version of teal (as of this issue) on commit 0065786, and running "tl check min_heap.tl" on [this file](https://gist.github.com/svermeulen/d0892dd6d4d3932afcdc1b98c3342edf), causes a crash with the following error:

```
tl.lua:7542: attempt to index local 't' (a nil value)
stack traceback:
	tl.lua:7542: in function 'resolve'
	tl.lua:7689: in function 'resolve'
	tl.lua:7684: in function 'resolve'
	tl.lua:7622: in function 'resolve'
	tl.lua:7647: in function 'resolve'
	tl.lua:7697: in function 'typevar_resolver'
	tl.lua:7761: in function 'resolve_typevars_at'
	tl.lua:8038: in function 'match_typevals'
	tl.lua:8093: in function 'to_structural'
	tl.lua:12253: in function 'cbkind_after'
	tl.lua:4980: in function 'recurse'
	...
	tl.lua:4765: in function 'fn'
	tl.lua:4972: in function 'recurse_node'
	tl.lua:13250: in function 'check'
	tl.lua:13386: in function 'process'
```

I noticed that I can get it to not crash by checking for nil here on line 7689:

```
if t.display_type ~= nil then
   copy.display_type, same = resolve(t.display_type, same)
end
```

hishamhm:
> I noticed that I can get it to not crash by checking for nil here on line 7689:

I believe this is the right answer! Mind sending a PR? :)

PR #828, opened 2024-10-16
Fix nil reference exception on issue #824

svermeulen:


svermeulen:
Fixes #824

hishamhm:
thank you!!

PR #832, opened 2024-10-18
fix: interface instances can be required concretely

### LINKED ISSUE #825: Typing a table as an interface makes it abstract
Minimal test case:
```lua
--mod1
local interface test1
    a: string
end

local mod1 = {} as test1
mod1.a = "foo"

return mod1
--mod2
local mod1 = require("mod1") -- interfaces are abstract; consider using a concrete record
```
In this example there's no reason to use an interface but in a situation where you need to be able to use an interface both for subtyping and typing (e.g. define a class's type and also use it for subtypes in inheritance, but also getting classes from a function call (replacing mod1 the table)) you'd have to have a seperate record which would be messy and bloated

Unless there's another way to do this (like function generics or something) could this be looked into? Thanks

### LINKED ISSUE #829: Placing interfaces in an included file produces error
The following works fine:

```
local interface IFoo
   bar: function(self)
end

local record Foo is IFoo
end

function Foo:bar()
   print("bar")
end
```

But when placing IFoo in a file like this:

```
local interface IFoo
   bar: function(self)
end

return IFoo
```

```
local IFoo = require("main.ifoo")

local record Foo is IFoo
end

function Foo:bar()
   print("bar")
end
```
We get error 'interfaces are abstract; consider using a concrete record' on the require line

hishamhm:
Fixes #825.
Fixes #829.

### DIFF
--- a/spec/lang/code_gen/local_type_spec.lua
+++ b/spec/lang/code_gen/local_type_spec.lua
@@ -239,4 +239,43 @@ describe("local type code generation", function()
       ]])
    end)
 
+   it("always elides local type require used as a variable, even if incorrect use of interfaces or aliases", util.gen([[
+      local interface IFoo
+      end
+
+      local type Alias = IFoo
+
+      local record Foo is IFoo
+      end
+
+      local function register(_id:any, _value:any)
+      end
+
+      local foo:Foo
+
+      register(IFoo, foo)
+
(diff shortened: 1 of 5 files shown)


hishamhm:
Since there has been no further feedback on this item, I am inclined to merge this as is.

Issue #841, opened 2024-10-27
io.input and io.output can also take a string as its argument


FractalU:
Since I updated `tl` from 0.15.3 to 0.24.1 I noticed that `tl` errors out if either `io.input` or `io.output` has a filename with the type string as its argument.

For example, the following code
```
print(io.input("text.txt"):read("a"))
```
produces the error message
```
========================================
1 error:
input.tl:1:16: argument 1: got string "text.txt", expected FILE
----------------------------------------
1 error
```
In Lua 5.4 and Luajit this code works fine. In the Lua 5.4 reference manual it's clearly written that it can also take a filename as its first argument. Below are the `io.input` and `io.output` sections of the reference manual.
```
io.input ([file])

When called with a file name, it opens the named file (in text mode), and sets its handle as the default input file. When called with a file handle, it simply sets this file handle as the default input file. When called without arguments, it returns the current default input file.

In case of errors this function raises the error, instead of returning an error code. 
```
and
```
io.output ([file])

Similar to [io.input](https://www.lua.org/manual/5.4/manual.html#pdf-io.input), but operates over the default output file.
```
This issue seems rather easy to fix. I already have a proposed solution. In the type signature of `io.input` and `io.output` change the types of the function argument from `? FILE` to `? FILE | string`.

In `tl.tl`  at line 139
```
      input: function(? FILE): FILE
```
and at line 149
```
      output: function(? FILE): FILE
```
Change these to
```
      input: function(? FILE | string): FILE
```
and at line 149
```
      output: function(? FILE | string): FILE
```
Btw. I also checked the `io.input` and `io.output` type signature of `tl` 0.15.3. There I see that the type of the function argument is  an option of the union of string and nominal file which is more or less the same as my proposal. Below are the lines in which the types of `io.input` and `io.output` are being defined in `tl` 0.15.3.

In `tl.tl` (version 0.15.3) at line 5402
```
            ["input"] = a_type { typename = "function", args = TUPLE { OPT(UNION { STRING, NOMINAL_FILE }) }, rets = TUPLE { NOMINAL_FILE } },
```
and at line 5409
```
            ["output"] = a_type { typename = "function", args = TUPLE { OPT(UNION { STRING, NOMINAL_FILE }) }, rets = TUPLE { NOMINAL_FILE } },
```

hishamhm:
Thanks for the report and investigation! Would you care to send a pull request with your fix? Thanks!!

FractalU:
[...]

PR #842, opened 2024-10-28
fix io.input and io.output type definitions

FractalU:
fixes https://github.com/teal-language/tl/issues/841


hishamhm:
@FractalU thank you!! It's always good to see the list of contributors grow :)

https://github.com/teal-language/tl/graphs/contributors

PR #849, opened 2024-11-04
fix: string.find type defintions

FourierTransformer:
Updating string.find type definitions to be variadic again.

As discussed here: https://github.com/teal-language/tl/discussions/847#discussioncomment-11147760


hishamhm:
Thank you!!

PR #869, opened 2024-12-07
Implement unread variable warning

euclidianAce:


hishamhm:
Love this — this is one of the few things from `luacheck` that I was still missing in `tl`! It's 4am here now, so I'll give this a closer look when I'm more awake before merging it :)

Issue #743, opened 2024-04-14
Allow indexing (disable type checking) on `any`


RobertBouillon:
Disable type-checking on the `any` type, or provide another type which allows Lua-style table traversal (`dynamic` ?)

### Use-Case
Sometimes the application provides a complex hierarchy we don't want to describe with Teal records. `any` type restricts access and the `table` type only provides access one level deep.

```lua
local function foo(bar:any)
  local foobar = bar.foo.foobar -- Can't index bar
end

local function foo(bar:table)
  local foobar = bar.foo.foobar -- Can't index foo
end

```

uramer:
I support this. It's very annoying to have to provide types even for things you don't care to type check for. Especially when the types you'd declare are not source of truth anyway because you are using third party code. Maybe an optional warning for traversing such "any" or "unknown" types would be useful as well

hishamhm:
Thank you for the feedback!

A behavior similar (but not identical) to what you asked for currently exists in the `unknown` type, which is the implicit type of variables that are not annotated. When a file has a `.lua` extension, the compiler goes into "lax" mode, and anything about `unknown` variables is ignored, other than reporting their existence as warnings. 

The idea was that one would take their `.lua` file, start running `tl check` on it, get a zillion `unknown` warnings, and start type-annotating them gradually, to the point where the file is fully annotated, then it becomes a `.tl` file, which can then be checked in the default "strict" mode. Of course, `.lua` files with partial annotations are not proper Lua files, so that was meant as a temporary state. But at least that makes the conversion process less grating than starting with a zillion errors (because in "strict" mode the errors due to insufficient annotation propagate a lot more).

What I didn't want is to be too lenient on missing annotations, to avoid the situation you see in TypeScript where people abuse `any` in places that could be typed trivially, so dissuading people from reaching out to `any` all the time is a constant concern in the TS world.

Note that in Teal, the `any` type has semantics that match TypeScript's `unknown` type: the only thing you can do with it is pass it around, or cast it. Effectively, the names `any` and `unknown` have opposite meanings in TS and Teal, with the difference that currently in Teal you cannot explicitly declare a variable as `unknown`. Another problem in TypeScript is that people often use TS `any` (their "do anything with it" type) in places where they really intended to use TS `unknown` (their "just pass it around" type). Teal started cautiously, by only exposing the "just pass it around" type.

[...]

PR #874, opened 2024-12-14
Fix/syntax error calling multiline call return value

### LINKED ISSUE #865: Teal unable to require base64
### What?
If trying to `tl run` a `.tl` file that requires the `base64` lua package, the command fails, with the error:
```
/usr/bin/lua5.4: /usr/local/share/lua/5.4/tl.lua:13477: /usr/local/share/lua/5.4/base64.lua:29:2: syntax error hint: construct starting here is not aligned with its 'end' at /usr/local/share/lua/5.4/base64.lua:65:1:
stack traceback:
        [C]: in function 'error'
        /usr/local/share/lua/5.4/tl.lua:13477: in local 'a_loader'
        /usr/local/share/lua/5.4/luarocks/loader.lua:104: in function </usr/local/share/lua/5.4/luarocks/loader.lua:101>
        (...tail calls...)
        [C]: in function 'require'
        main.tl:1: in main chunk
        (...tail calls...)
        /usr/local/lib/luarocks/rocks-5.4/tl/0.24.1-1/bin/tl:977: in main chunk
        [C]: in ?
```
This worked on a previous version of teal (0.15.3-1), but since updating to 0.24.1, does not work.
It seems like teal is trying to validate the syntax on the `base64.lua` file, but (unless I'm misunderstanding, which is very possible), it should only check the types and syntax on the `base64.d.tl` file, and should just load `base64.lua` like a standard lua file.

### Minimal example
```lua
-- main.tl
require("base64")

-- base64.d.tl
local record base64 end
return base64
```

```sh
luarocks install base64
tl run main.tl
```

JR-Mitchell:
### What?
This PR introduces the `ParseLang` enum and `parse_lang` field to the teal parser.
This allows different behaviour when parsing depending on whether a file is a `.lua` file, a `.tl` file, ~~or a `.lax.tl` file (more on this below).~~
This is then used in this PR to ensure that lua code which calls functions on a new line, or after a multiline string literal, parse correctly.

### Why?
Primarily, as a proposal to resolve #865.

**Language syntax differences**
#748 added logic to the parser which creates a difference in parsing behaviour between *Lua* and *Teal*.

In *Lua*, the following is valid syntax:
```lua
print
("hello world")
```
and results in "hello world" being printed when run.

However, in *Teal*, this new casting syntax is permitted:
```lua
(myVal as MyType):method()
```
Which introduces situations where it is not possible to determine whether a statement like this is intended as a new statement, or a call to a function or method on a previous line, such as:
```lua
local record MyType
    getComparator: function(self: MyType): function(other: MyType): MyType
end

local a = {}
local b = {}
local c = (a as MyType):getComparator()
(b as MyType):getComparator()
```
The last 2 lines here could either be interpreted as
```lua
local c = ((a as MyType):getComparator()(b as MyType)):getComparator()
```
or as
```lua
local c = (a as MyType):getComparator();
(b as MyType):getComparator()
```

#748 resolves this ambiguity by making it so that *Teal* does not allow function calls on new lines.
However, this meant that the teal parser was no longer compatible with the parsing of *Lua*, meaning that if any *Teal* code required any *Lua* code that used this syntax, it would fail.
For example, this broke the `base64` *Lua* library in teal.

This PR aims to handle that the parser is used to parse 2 different languages which now have subtly different parsing rules by passing a flag in the `ParseState` for which language's parsing rules are being used, so that *Teal*- or *Lua*-specific behaviour can be conditionally applied when parsing.

~~__Migration of *Lua* code to *Teal*__~~

~~This PR also introduces a third `ParseLang`, `lax-tl`, representing files with a filename ending `.lax.tl`.~~
~~This is my proposed approach to resolving some of the comments mentioned in #743 about gradual migration of *Teal* code to *Lua* code - in particular:~~
~~> Of course, .lua files with partial annotations are not proper Lua files~~

~~and~~
~~> Requiring .lua files into .tl files practically compels you to rewrite the code unless you write a .d.tl file for every .lua file and maintain them alongside your .lua code~~

~~This PR allows for `.lua` files to be renamed to `.lax.tl` files, which will now be parsed with the *Teal* syntax as opposed to *Lua* syntax, but will have lax type checking performed on them (as if they were `.lua` files).~~
~~This suggested approach would hopefully make it easier to migrate *Lua* code to *Teal*, as you simply rename your `.lua` files to `.lax.tl` files and begin to add type annotations, and then once they are fully migrated, drop the `.lax` from the filename, as opposed to the previous approach, which would require an additional `.d.tl` file and also leave you with a `.lua` file that isn't actually valid *Lua*.~~

### Other notes
I made the `parse_lang` parameter optional, primarily for backwards compatibility (so that any use of functions such as `tl.parse_program` would not break due to the extra added argument), and also because there is one situation where I wasn't sure what to pass (line 7012 in `tl.tl`).
However, I am aware that making this non-optional may lead to more robust code, as the type checker would ensure that a non-nil argument is always passed for it.

Please let me know in general whether you think this PR is a good or bad approach to resolving this issue, and specifically whether you think I should rewrite so that `parse_lang` is non-optional.


pbdot:
Instead of adding a file extension, which then tooling also needs to be made aware of, perhaps it would be cleaner to use something like a compiler annotation in the Lua source, which the tl picks up on.  Annotations could be used for other purposes as well, including existing compiler options.  For example, https://typescripttolua.github.io/docs/advanced/compiler-annotations

JR-Mitchell:
Yeah, that's a fair point. I'll rework this PR to remove the `.lax.tl` stuff I added in the next couple of days.

tl.tl
```diff
@@ -6998,6 +7008,7 @@ local function add_compat_entries(program: Node, used_set: {string: boolean}, ge
    local function load_code(name: string, text: string)
       local code: Node = compat_code_cache[name]
       if not code then
+         -- TODO: does this need a parse_lang?
```
hishamhm, line 7011: I don't think this _needs_ it because we only load well-known and well-behaved code here (the compatibility chunks listed below in `add_compat_entries`), but since these are Lua chunks, it could be a good idea to explicitly pass `"lua"` as the `parse_lang`, if only for documentation sake.

hishamhm:
@JR-Mitchell it's late here so I won't trust my own review at 1:30am and will take a closer look at this in the coming days,  but overall it's looking good!

Regarding the discussion with @pbdot above, I wonder if the way to go would be to add a `#pragma lax on` like we do for `arity`. I've avoided so far it for the sake of keeping the `.lua` and `.tl` boundaries more separate and encouraging people to just go all the way with typing, but if the current state of things makes it _more_ difficult to migrate, then perhaps the pragma would be the best bridge?

hishamhm (approved):
looking good! I think we can just remove that TODO (by either keeping `parse_lang` implicit or explicit there, your call!)

JR-Mitchell:
@hishamhm thanks!
I've updated so that `parse_lang` is explicitly `"lua"` in that case, I think that makes the code overall clearer.
I also made the `parse_lang` non-optional in `default_env_opts`, since it was explicitly passed at all the call sites anyway, and probably should be specified whenever used.

JR-Mitchell:
Btw I don't have permissions to merge to this repo, so someone else will have to do that for me if they're happy to

hishamhm:
Will do when I'm back at it -- I'm taking a holiday break but will get back to it early Jan

hishamhm:
@JR-Mitchell I squashed some of the commits and merged the PR manually! Thank you!!

PR #890, opened 2025-01-02
generalize internal representation of generic types

### LINKED ISSUE #787: cannot match types if union is behind a type alias
```lua
local type nilable<T> = T | nil
local function divide(a: number, b: number): nilable<number>, nilable<string>
    if b == 0 then
        return nil, "division by zero"
    end
    return a / b, nil
end

local x, err = divide(10, 2)
if err is nil and not x is nil then
    print("Result: "..x)
else
    print("Error: "..err)
end

```

Results in `x` being inferred as `nil`, and `err` being inferred as `nilable<string>`. I am using the `next` branch.

### LINKED ISSUE #880: Explicit type annotation for generics causes `X<T> is not a X<T>` error
In situations where one generic type is defined in terms of another, sometimes explicit type indications with the second generic cause the typechecker to fail with the error message `X<T> is not a X<T>`, but leaving the type implicit causes no errors.

Minimal example:
```lua
local record TypeA<T> end

local type TypeB<T> = {TypeA<T>}

-- this works:
local function TypeA_new<T>(): TypeA<T>
  local a: TypeA<T> = {}
  return a
end

-- this fails: "in return value: TypeB<T> is not a TypeB<T>"
local function TypeB_new<T>(): TypeB<T>
  local a: TypeA<T> = {}
  local b: TypeB<T> = {a}
  return b
end

-- but it works when leaving the type of b implicit:
local function TypeB_new_<T>(): TypeB<T>
  local a: TypeA<T> = {}
  local b = {a}
  return b
end
```

hishamhm:
All types that have type variables are now represented as a GenericType record, which holds a non-generic Type and an array of type arguments.

This change is because originally we only cared about generic records and generic functions, but once we have the `local type MyType<T> = ...` syntax, other types can be generic as well (in particular, unions).

Instead of replicating generic support logic in the implementation of each type, we factor it out into a type-level term which encapsulates the application of type variables, which is something more like second-order lambda calculus.

(See https://en.wikipedia.org/wiki/System_F and the ensuing rabbit hole.)

Fixes #880.
Fixes #787.

### DIFF
--- a/spec/cli/types_spec.lua
+++ b/spec/cli/types_spec.lua
@@ -377,7 +377,7 @@ describe("tl types works like check", function()
          assert.same({
             ["17"] = 7,
             ["20"] = 2,
-            ["25"] = 10,
+            ["25"] = 17,
             ["31"] = 9,
          }, by_pos["2"])
       end)
--- a/spec/lang/call/generic_function_spec.lua
+++ b/spec/lang/call/generic_function_spec.lua
@@ -424,6 +424,19 @@ describe("generic function", function()
       { y = 6, msg = "cannot infer declaration type; an explicit type annotation is necessary" },
    }))
 
+   it("can use record type variable in record function", util.check_type_error([[
+      local record Container<T>
+         try_resolve: function(Container):T
+      end
+
+      function Container:resolve<U>():U
+         local t = self:try_resolve()
+         return t
+      end
+   ]], {
+      { y = 7, msg = "got T, expected U" },
+   }))
+
    it("works when an annotation is given", util.check([[
       local record Container
          try_resolve: function<T>(Container):T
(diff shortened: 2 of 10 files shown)


hishamhm:
I just saw two issues already when running the LuaRocks codebase through it: 

modules that return a generic type

```
./luarocks/search.tl:16:55: missing type arguments in generic<K> table.SortFunction<K> | Ordering<K>
```

https://github.com/luarocks/luarocks/blob/master/src/luarocks/search.tl#L16

and a stricter resolution of type arguments (which I need to look closer, but might be a legitimate error that it has now caught)

```
./luarocks/cmd/write_rockspec.tl:415:81: argument 3: got Ordering<string>, expected table.SortFunction<number | string> | Ordering<number | string>
```

https://github.com/luarocks/luarocks/blob/master/src/luarocks/search.tl#L16

hishamhm:
Error 1 there was being misreported (it was being reported at the wrong location). It was a legitimate error, a missing `<K>`.

Error 2 is a bit trickier because it is a type check that we would _like_ to accept (because we know that `Ordering<string>` is a valid `Ordering<string | number>`, but only because we know how the Ordering object works), but we can't, because the compiler has no way to tell that values of the type variable's type are only used for reading and not for writing... typical covariance/contravariance limitation. Currently, type variables are invariant, for simplicity. The easy fix is to change the field to `Ordering<string | number>` in the LuaRocks codebase.

hishamhm:
All right, I decided to do the brave thing and merge this already.

The fixes contained here are important enough so that it's clear that this should be the baseline moving forward. If this has introduced any further regressions, then let's fix it in the main branch.

PR #894, opened 2025-01-03
make `v is T and v or e` imply not-T for expression e

### LINKED ISSUE #878: `is` should imply truthiness for non-boolean non-nil types
(Possibly related to/duplicate of #776, or maybe a special case of it?)

Consider the following:

```lua
local record R end

local function convert(_: string): R end

local u: string | R
local _r: R = u is R
	and u
	or convert(u)
```

```
$ tl check example.tl
example.tl:8:13: argument 1: got string | R, expected string
```

The only case where the argument to `convert` will not be a string is when `u` itself is nil. But since `u is R and u` (-> `type(u) == "table" and u`) will never produce a `nil`, `convert`'s argument is a `string` (well, `string | nil`).

In fact this error can be avoided by using `and assert(u)` instead of just `and u`.

I haven't dug around much in the fact inference code so I'm not sure how feasible it is to say a given variable should be truthy in a given context. Also with custom `is` implementations it may not be as simple as this, but it would be nice to not need a runtime assert for this.

hishamhm:
Only for "truthy types" (i.e. all except boolean and nil).

Closes #878.

### DIFF
--- a/spec/lang/inference/or_spec.lua
+++ b/spec/lang/inference/or_spec.lua
@@ -1,0 +1,15 @@
+
+local util = require("spec.util")
+
+describe("inference in 'or' expressions", function()
+   it("`v is T and v or _` for a truthy T infers that _ is of type not-T (#878)", util.check([[
+      local record R end
+
+      local function convert(_: string): R end
+
+      local u: string | R
+      local _r: R = u is R
+         and u
+         or convert(u)
+   ]]))
+end)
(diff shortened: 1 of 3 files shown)


hishamhm:
@euclidianAce what do you think? worthy or too hacky?

euclidianAce:
Looks good to me! :+1:

> worthy or too hacky?

I think small special cases like this are worth it to support some existing Lua idioms.

hishamhm:
Merged — hopefully we'll be able to remove this once we handle nils separately.

PR #895, opened 2025-01-03
make function types as constraints usable

hishamhm:
See #211.


hishamhm:
This does not fully address all the things mentioned in #211. I don't think it provides much improvement over just using the `F` type directly (in the test case example), but it makes the behavior of type constraints more consistent.

PR #899, opened 2025-01-08
gen: change behavior to check by default, add --no-check

hishamhm:


tl
```diff
@@ -416,7 +416,8 @@ local function get_args_parser()
 
    local gen_command = parser:command("gen", "Generate a Lua file for one or more Teal files.")
    gen_command:argument("file", "The Teal source file."):args("+")
-   gen_command:flag("-c --check", "Type check and fail on type errors.")
+   gen_command:flag("-c --check", "Type check and fail on type errors. This is the default."):hidden(true)
```
mtdowling, line 419: Is it possible to continue to support this argument but just not document in --help?

tl
```diff
@@ -416,7 +416,8 @@ local function get_args_parser()
...
+   gen_command:flag("--no-check", "Do not fail on type errors, only on syntax errors.")
```
mtdowling, line 420: If you keep --check in the help output, it may be worth validating that both --check and --no-check aren't provided at the same time. If you removed --check from help output, then I wouldn't worry about validation.

tl
hishamhm: that is what `hidden(true)` does!

PR #906, opened 2025-01-13
fix: do not over-fresh type variables

### LINKED ISSUE #905: Regression: Generic error "got T, expected T"
Hard to explain this one.  I tried reducing it down as much as possible.

The following compiles fine for previous versions of teal.  For current version (master) we get error `main.tl:11:26: argument 1: argument 1: got T, expected T`:

```
local record Mapper<T1>
   map:function<T2>(self, selector:(function(T1):T2))
end

local function to_mapper<T>(_values:{T}):Mapper<T>
   error("unimplemented")
end

local function _run<T>(values:{T})
   to_mapper(values):map(function(_:T):integer return 5 end)
end

_run({
   "five"
})
```

hishamhm:
Fixes #905.

### DIFF
--- a/spec/lang/call/generic_function_spec.lua
+++ b/spec/lang/call/generic_function_spec.lua
@@ -591,5 +591,23 @@ describe("generic function", function()
          a.f()
       end
    ]]))
+
+   it("should not over-fresh type variables (regression test for #905)", util.check([[
+      local record Mapper<T1>
+         map:function<T2>(self, selector:(function(T1):T2))
+      end
+
+      local function to_mapper<T>(_values:{T}):Mapper<T>
+         error("unimplemented")
+      end
+
+      local function _run<X>(values:{X})
+         to_mapper(values):map(function(_:X):integer return 5 end)
+      end
+
+      _run({
+         "five"
+      })
+   ]]))
 end)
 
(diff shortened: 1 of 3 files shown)


hishamhm:
@svermeulen @mtdowling Testing of this PR with your codebases is appreciated!

mtdowling:
I'm happy to test this. Is there a way to have Luarocks install a specific branch of tl? You taught me about --dev the other day, but not sure about how to use a branch.

catwell:
@mtdowling You can add `branch = "..."` in the `source` table of the rockspec, but for Teal it is probably easier to clone the repository, checkout the branch you want and run `tl` directly from there.

hishamhm:
`luarocks build --dev --branch fix-906 tl`  should work too!

mtdowling:
I _think_ I tested this and works for my project.

For future reference for my future self, I did this:

1. Cloned this repo and checked out this branch.
2. Ran `luarocks build` in the `tl` directory.
3. Verified that the tl.lua file in `/opt/homebrew/share/lua/5.4/tl.lua` has this change.
4. Verified that cyan installed via homebrew uses this same path.
5. Ran cyan to build my project successfully.
6. Double checked that cyan uses this file by making it error out, and it did... so my assumption was correct that it was used :)

hishamhm:
As I half-suspected, it looks like this is not the right solution; it causes #909. Back to the drawing board.

hishamhm:
Replacing this with #911 which looks like a more robust fix!

PR #917, opened 2025-01-19
Add Helix editor on README.md

Andre-LA:
Starting from 25.01, Helix now supports Teal built-in, the editor also integrates LSP by invoking teal-language-server from PATH.

I was in doubt about the URL since the configuration it's not a plugin (only the LSP needs to be installed and available on PATH), so I decided to point to the documentation on language support.

The URL uses the ["Text Fragments"](https://wicg.github.io/scroll-to-text-fragment/) feature (which is supported on major browsers) to highlight (and jump) to the Teal entry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ https://github.com/teal-language/teal-types — check it out and make your contr
 
 ## Text editor support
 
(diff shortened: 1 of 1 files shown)


hishamhm:
Thank you!!

PR #923, opened 2025-01-29
Escape percentage signs for some more fields

mbartlett21:
Ran in to this a couple of times with some parsing libraries.


hishamhm:
I guess at some point we'll need something a little less ad-hoc to properly escape those keys!

PR #924, opened 2025-01-31
fix close signature

fperrad:
quotes from Lua manual:

    When closing a file handle created with io.popen, file:close returns the same values returned by os.execute.


hishamhm:
Thanks!!

PR #940, opened 2025-03-21
Fixes for `...` in macroexp

### LINKED ISSUE #939: `macroexp` and `...`
Macro functions that take varargs and then pass them to another function don't get compiled properly.

E.g:

```lua
local macroexp runformat(s: string, ...: any): string
   return string.format(s, ...)
end

print(runformat('%s %s %s', 'varargs', 'dis', 'appear'))
```

Compiles to

```lua
print(string.format('%s %s %s', ...))
```

mbartlett21:
Fixes #939 

This does a couple of things. First, it adds '...' to the elements that get iterated over, so it warns when there is multiple use of varargs. It then codegens `...` as the leftover arguments.

### DIFF
--- a/spec/lang/code_gen/macroexp_spec.lua
+++ b/spec/lang/code_gen/macroexp_spec.lua
@@ -194,5 +194,19 @@ describe("macroexp code generation", function()
          print("the 'x' fields are equal")
       end
    ]]))
+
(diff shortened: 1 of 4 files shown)


hishamhm:
@mbartlett21 Thank you, looks great!!

PR #946, opened 2025-04-10
Add parsing for various string functions

mbartlett21:
This adds some handling for literals in:
- `string.format`
	- With the parsing, I've left it at the first stage of how Lua does it without validating the modifiers. As far as I know it should accept anything that Lua accepts.
- `string.{pack, unpack}`
    - (I could probably validate the string if it is in string.packsize, but that won't affect arguments or returns)
    - I haven't allowed numeric arguments to packing in 'z', 's' or 'c', as they always get unpacked as a string
- `string.{find, match, gsub, gmatch}`
    - I think this should be fairly complete.

In order to handle it better, I've added override arguments to the functions for validation.

I've also added a (unused) strict mode to the pattern parsing that complains when the pattern appears to be relying on implementation details (like a `*` or other repetition after a capture always being literal)

### DIFF
--- a/tl.tl
+++ b/tl.tl
@@ -9941,9 +9941,9 @@ do
          "type_dot" -- a dot-call where the receiver is a type, e.g. `MyRecord.my_func()`
       end
 
-      local check_call: function(self: TypeChecker, w: Where, wargs: {Where}, f: FunctionType, args: TupleType, expected_rets: TupleType, cm: CallMode, argdelta: integer): boolean, {Error}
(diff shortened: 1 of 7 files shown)


mbartlett21:
Lua 5.1 doesn't have goto :(

tl.tl
```diff
@@ -11333,28 +11641,301 @@ do
...
+         end
+
+         return (self:type_check_function_call(node, a, b, argdelta, args, nil))
+      end,
+   }
+
+   function TypeChecker:type_check_funcall(node: Node, a: Type, b: TupleType, argdelta?: integer): InvalidOrTupleType
+      argdelta = argdelta or 0
+
+      local special_tyck = special_functions[a.special_function_handler]
```
hishamhm, line 11927: Looks like the handling of `special_tyck` could move inside `type_check_function_call`, then it only needs to be handled once.

hishamhm: Ah, nevermind, I see now why not!

PR #955, opened 2025-04-25
Resolve string metatable with modules

### LINKED ISSUE #933: shadowing string breaks std string metatable detection
tst.tl
```teal
local string = "text"
print(("%i"):format(42))
```
 
```
# tl check tst.tl:
 1 warning:                                                                                                                                                                   
tst.tl:1:7: variable shadows previous declaration of 'string'                                                                                                                
----------------------------------------                                                                                                                                     
1 warning                                                                                                                                                                    
1 error:                                                                                                                                                                     
tst.tl:2:14: cannot index key 'format' in type string "%i"                                                                                                                   
----------------------------------------                                                                                                                                     
1 error
```

Same error without warning produced if I store string from standard library to local variable
```
local string = string
print(("%i"):format(42))
```

mbartlett21:
This results in it no longer caring about `local string` to resolve it.
Fixes #933

### DIFF
--- a/spec/lang/stdlib/string_spec.lua
+++ b/spec/lang/stdlib/string_spec.lua
@@ -1,6 +1,12 @@
 local util = require("spec.util")
 
 describe("string", function()
+   describe("literal", function()
+      it("doesn't care about local string", util.check([[
+         local string = "text"
+         print(("%i"):format(42))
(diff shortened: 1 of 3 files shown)


hishamhm:
Nice solution!

Issue #945, opened 2025-04-09
Some kind of retained state issue with generic interfaces


mtdowling:
I'm trying to define a typed state machine, but I'm running into an issue where the compiler seems to be retaining state after parsing a previous type (or something like that). This is only happening when I have a generic function the StateMachine type.

```lua
local interface StateValue
    value: string
end

local interface StateMachine<S is StateValue>
    showTheBug: function(self, hi: S): S
    -- ^ comment this out and it works
end

local function newStateMachine<S is StateValue, M is StateMachine<S>>(stateMachine: M)
end

local record Editor is StateMachine<Editor.On | Editor.Off>
    record On is StateValue where self.value == "on" end
    record Off is StateValue where self.value == "off" end
end

local record App is StateMachine<App.Game | App.Menu>
    record Game is StateValue where self.value == "game" end
    record Menu is StateValue where self.value == "menu" end
end

newStateMachine(Editor) -- Works for some reason.

newStateMachine(App) -- < Doesn't work
-- ^ argument 1: given type App does not satisfy StateMachine<S> constraint in type variable M
```

Also note that if I move the method call using App above Editor, then the Editor call fails.

```lua
newStateMachine(App)
newStateMachine(Editor) -- < Doesn't work
```

BLKTower:
I believe this is the same issue as [931](https://github.com/teal-language/tl/issues/931).

mtdowling:
Closing as a duplicate of #931

PR #962, opened 2025-05-02
fix: resolve constraints when freshening typevars

### LINKED ISSUE #931: Bug: using `interface` with `generic` cause type resolve incorrectly
```lua
local interface I<L>
    _: L
end
local record A end
local record IA is I<A> end
local record B end
local record IB is I<B> end

local function Create<L, C is I<L>>(l: L, c:C): C
    return {}
end

local function Test(a: A, b: B)
    -- Test 1
    Create(a, IA)
    Create(b, IB) -- Error: argument 2: given type IB does not satisfy I<L> constraint in type variable C
    Create(b, IA) -- No error, but it should.
    -- Test 2 -- No error if separate 2 function calls in different scope
    -- do
    --     Create(a, IA)
    -- end
    -- do
    --     Create(b, IB)
    -- end    
end
```
It seems after `Create` first gets called, all the following function calls in the same scope will use the same type resolve results.
Desired behavior: the teal compiler can resolve the type correctly.

hishamhm:
Fixes #931.


mtdowling:
Works for https://github.com/teal-language/tl/issues/945 and also works with my codebase. Thanks!

PR #963, opened 2025-05-03
fix: macroexps work with generics

### LINKED ISSUE #941: macroexp with generics?
I have a kind of typed bag like this:

```lua
record Archetype
    columns: {Component:{Component}}
end
```

Where really each entry in the columns map is a `T -> {T}`, but I can't express that in the type system. You can't do this lots of languages either, like Java for example.

```lua
local archetype = getArchetype()
local sprockets = archetype.columns[Sprocket] as {Sprocket} -- < I want to remove this cast
```

So right now, I added a function to get the component, like this:

```lua
record Archetype
    columns: {Component:{Component}}
    getColumn: function<T is Component>(self, component: T): {T}
end
```

This function just exists to make things type correctly:

```lua
function Archetype:getColumn<T is Component>(component: T): {T}
    return self.columns[component] as {T}
end
```

I'd like to make this a `macroexp` to make it a zero cost abstraction, but `macroexp` doesn't support generics at the moment.

```lua
record Archetype
    columns: {Component:{Component}}
    getColumn: function<T is Component>(self, component: T): {T} = macroexp(self: Archetype, component: T): {T}
        return self.columns[component] as {T}
    end
end
```

How feasible is it to support generics in `macroexp`? Is this opening up a world of problems and edge cases, or is it just something that hasn't been implemented yet and should be straightforward to add?

hishamhm:
Fixes #941.

### DIFF
--- a/spec/lang/macroexp/generics_spec.lua
+++ b/spec/lang/macroexp/generics_spec.lua
@@ -1,0 +1,79 @@
+local util = require("spec.util")
+
+describe("macroexp with generics", function()
+   it("can resolve type arguments", util.gen([[
+      local interface Component
+         name: string
+      end
+
+      local record Sprocket is Component
+         where self.name == "sprocket"
+         size: integer
+      end
+         function Sprocket:new(s: integer): self
+         return { name = "widget", size = s }
+      end
+
+      local record Widget is Component
+         where self.name == "widget"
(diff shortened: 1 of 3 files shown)


mtdowling (commented):
LGTM!

PR #969, opened 2025-05-10
extras/tl.usage: add shell completions for `tl`

ildar:
First build with `complgen` then "source" the output file. E.g.
$ complgen --bash _tl extras/tl.usage
$ source _tl

### DIFF
--- a/extras/tl.usage
+++ b/extras/tl.usage
@@ -1,0 +1,30 @@
+# Compile this file like so:
+#   complgen --bash _tl tl.usage
+
+tl [<GLOBAL-OPTIONS>]... check                 "Type-check one or more Teal files." <PATH>...;
+tl [<GLOBAL-OPTIONS>]... gen                   "Generate a Lua file for one or more Teal files." <PATH>...;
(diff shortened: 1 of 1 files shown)


ildar:
example generated bash completion: 
[_tl.txt](https://github.com/user-attachments/files/20138282/_tl.txt)

0komo:
You don't practically need this, argparse already has an option to create completion for the command. You only need to append `:add_complete_command()` after `argparse(...)`. Unless this `complgen` supports variety of shells besides Bash, Zsh, and Fish.

ildar:
Ah! then it's definitely better to go this path.
/close

ildar:
though I just checked and the option I offered is a little better: `--wdisable` and `--werror` have full completions, `argparse`'s don't.
JFYI. still don't think it should be merged.

0komo:
> though I just checked and the option I offered is a little better: `--wdisable` and `--werror` have full completions, `argparse`'s don't. JFYI. still don't think it should be merged.

That can be fixed tho, but will probably need some tweak to get it working.

PR #973, opened 2025-05-12
feat(cli): add completion command

0komo:
- Related #969

### DIFF
--- a/tl
+++ b/tl
@@ -370,6 +370,8 @@ end
 local function get_args_parser()
    local parser = argparse("tl", "A minimalistic typed dialect of Lua.")
 
+   parser:add_complete_command()
+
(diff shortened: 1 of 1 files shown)


hishamhm:
@0komo Thanks! Could you add some small test under `spec/cli` ?

hishamhm:
the change itself LGTM! For the test, you can take some other file under spec/cli such as the one for warnings and make a similar one, with a small test that triggers the command and then checks some of the output. It doesn't need to be exhaustive, a basic sanity check is fine!

0komo:
@hishamhm Here you go, you may review it if you want.

hishamhm:
All good! Thank you!!

PR #978, opened 2025-05-14
Allow using `-` to read/write stdin/stdout

mbartlett21:
This just adds an extra check for special-casing the filename.


tl
```diff
@@ -236,7 +236,9 @@ end
 
 local function process_module(filename, env)
    local module_name = filename_to_module_name(filename)
-   local result, err = tl.process(filename, env)
+   local fd
+   if filename == '-' then fd = io.input() end
+   local result, err = tl.process(filename, env, fd)
```
hishamhm, line 241: This will process standard input with a filename of `-` (for error messages, etc.)

Should we change `filename` to something like `"(standard input)"` here? What do other compilers do?

mbartlett21: gcc uses `<stdin>`

```shell
$ echo '#include "test.c"' | gcc -x c -
<stdin>:1:10: fatal error: test.c: No such file or directory
compilation terminated.
```

hishamhm: Lua itself calls it `stdin` too:

```
echo "if bug then then" | lua -
lua: stdin:1: unexpected symbol near 'then'
```

I like the idea of using some "special symbol" like angle brackets or parentheses to indicate that it's not a real filename (even though technically those _are_ valid filename characters). I have no preference here, up to you!

mbartlett21: Changed it to use `<stdin>` and `<stdout>` now.

mbartlett21:
I've also now made it default to using `stdout` when reading `stdin` (unless a particular output file is specified). This means that you can use `tl gen -c -q -` in a pipeline.

mbartlett21:
I've added some more options in the util module to have temporary files. Also a way to pipe to a command, since Lua's io.popen doesn't allow read and write.

I've just added extra tests under the quiet spec, seeing as using stdout isn't very helpful without `--quiet`

PR #979, opened 2025-05-15
Update the typechecking for `or` expressions

### LINKED ISSUE #975: Bug: string.format cannot infer type when used with 'or'
Testcase:

```lua
local function test(x ?: integer)
    print(string.format('%d', x))         -- OK
    print(string.format('%d', x or 0))    -- argument 2: got <any type> (inferred at :3:33), expected integer

    print(string.format('%d', (x or 0) as integer))  -- also OK, but requires type cast
end
```

Though type inference works as expected with `%s` string conversion specifier.

The `or` operator is quite idiomatic in these cases, so I think it should not cause an error.

mbartlett21:
Changes:
- If the types of an `or` statement are equal, then it evaluates to that type.
- If one of the types is a union and the other isn't (and it is convertible into it), then the resulting type is that union.
- If the types are each able to be converted into the other but they aren't equal, then it adds a warning and just uses the second type (this was previously accepted and always gave the second type).
- If the types aren't convertible into each other, it now accepts it if there is an expected type that they can be converted into (see the superclass test, which didn't pass before)

This should fix #975 .

### DIFF
--- a/spec/lang/operator/or_spec.lua
+++ b/spec/lang/operator/or_spec.lua
@@ -166,6 +166,7 @@ describe("or", function()
       wants_b(also_still_not_b)
    ]], {
       { y = 20, x = 15, msg = "got A | B, expected B" },
+      { y = 24, x = 15, msg = "got A | B, expected B" },
       { y = 27, x = 15, msg = "got A | B, expected B" },
    }))
 
--- a/spec/lang/operator/or_spec.lua
+++ b/spec/lang/operator/or_spec.lua
@@ -157,8 +157,16 @@ describe("or", function()
       local b = ab is A and ab.b or ab -- ab.b may be nil, causing the value of type A to be returned via 'or'
 
       wants_b(b)
+
+      local real_b: B = {tag="b"}
+      local also_not_b = ab or real_b
+      wants_b(also_not_b)
+
+      local also_still_not_b = real_b or ab
+      wants_b(also_still_not_b)
    ]], {
       { y = 20, x = 15, msg = "got A | B, expected B" },
+      { y = 27, x = 15, msg = "got A | B, expected B" },
    }))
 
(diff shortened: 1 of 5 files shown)


spec/lang/operator/or_spec.lua
```diff
@@ -191,6 +191,10 @@ describe("or", function()
       local _print_testcty: nil = _testc
       local _print_testdty: nil = _testd
    ]], {
+      { y = 6, msg = "the resulting type is ambiguous: A or B" },
```
hishamhm, line 194: Ohhh, I was wondering how we were ending up with an ambiguous type here -- in other words, I understand why `A :> B` (`rb is ra`) but it wasn't immediately clear to me why `B :> A`... 

Looks like the case `interface is record` is being matched _structurally_, which is odd! I don't remember why I ever put that in.

I just tried simply deleting the `["interface"]["record"]` rule from `subtype_relations`, and it actually picked up a couple of type declaration bugs in the compiler code (and the test suite passed otherwise).  I'll open a PR with that change!

hishamhm: Please check out #980 !

tl.tl
```diff
@@ -13755,22 +13755,42 @@ do
...
+                  if is_same then
+                     t = ua
+                  elseif (a_ge_b or b_ge_a) and not ambiguous then
+                     local larger_type = b_ge_a and ub or ua
+                     t = larger_type
+                  elseif expected and self:is_a(ua, expected) and self:is_a(ub, expected) then
+                     t = self:infer_at(node, expected)
+                  end
+
+                  if ambiguous and not t then
```
hishamhm, line 13787: If the fix in #980 is good, is it still possible to produce an ambiguous type like this?

mbartlett21:
Also superseded by
- #983

PR #980, opened 2025-05-15
fix: do not match interfaces to records structurally

hishamhm:
This rule was causing a situation where we could get a declarations `interface I` and `record R is I` resulting in `I :> R` and `R :> I` at the same time, depending on their contents. This produces ambiguities down the line: See https://github.com/teal-language/tl/pull/979/commits/a61a3c6234d1a0f8df742d88da2e14ce24ecfd22#r2091225922

I'm not sure why that rule ended up there in the first place.

The fact that removing that rule caught real type declaration bugs in the compiler is a good sign.

Perhaps I was assuming that checks between those types would happen earlier on at the nominal stage, or I added that line in a transitional stage in the introduction of interfaces.

I'm happy that the declarative style of the `subtype_relations` table made it easy to spot what was going on, though!

### DIFF
--- a/spec/lang/subtyping/interface_spec.lua
+++ b/spec/lang/subtyping/interface_spec.lua
@@ -218,4 +218,24 @@ describe("subtyping of interfaces:", function()
    ]], {
       { y = 12, msg = "'move' does not match definition in interface Animal" }
    }))
+
+   it("interface :> record but not the other way around", util.check_type_error([[
+      local interface I
+      end
+
+      local record R is I
+      end
+
+      local r: R
+      local i: I
+
+      local wants_i1: I = i or r
+      local wants_i2: I = r or i
+
+      local wants_r1: R = i or r
+      local wants_r2: R = r or i
+   ]], {
+      { y = 13, msg = "I is not a R" },
+      { y = 14, msg = "I is not a R" },
+   }))
 end)
--- a/tl.lua
+++ b/tl.lua
@@ -9467,7 +9467,6 @@ do
             return self:same_type(a, b)
          end,
          ["array"] = TypeChecker.subtype_array,
-         ["record"] = TypeChecker.subtype_record,
          ["tupletable"] = function(self, a, b)
             return self.subtype_relations["record"]["tupletable"](self, a, b)
          end,
--- a/tl.tl
+++ b/tl.tl
@@ -2580,7 +2580,7 @@ local function failskip(ps: ParseState, i: integer, msg: string, skip_fn: SkipFu
    return skip_i
 end
 
-local function parse_type_body(ps: ParseState, i: integer, istart: integer, node: Node, tn: BodyTypeName): integer, GenericType | StructuralType
+local function parse_type_body(ps: ParseState, i: integer, istart: integer, node: Node, tn: BodyTypeName): integer, GenericType | FirstOrderType
    local typeargs: {TypeArgType}
    local def: FirstOrderType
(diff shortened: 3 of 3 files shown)


mbartlett21:
There was something else I also noticed with the record subtyping. It seems to allow extra fields that don't exist in the record it is being converted into.

See for example the end of the tl.lex function. where it is magically converted into a Token type that doesn't have an `i` field.

mbartlett21:
I've done a partial rebase of my `or` changes onto this, but there are still some ambiguous types coming up. However, they seem to mostly be nominal vs resolved types. (which I have an individual commit on that at https://github.com/mbartlett21/tl/tree/nominal-comparison which I can merge into here if you would like)

There is a couple of other ambiguous types though that even extra field checking wouldn't help with:

```lua
local record A where self.field == "a"
	field: string
end
local record B where self.field == "b"
	field: string
end

local ab: A|B
local b: B

-- ambiguous: A can structurally match B, therefore A|B can match B
local c = ab or b
```

Perhaps it should be special casing where there is a `__is` metamethod?

hishamhm:
> However, they seem to mostly be nominal vs resolved types. (which I have an individual commit on that at https://github.com/mbartlett21/tl/tree/nominal-comparison which I can merge into here if you would like)

This looks okay, but without fully thinking it through, I'm a bit wary wondering if it slides us further towards structural typing.

> -- ambiguous: A can structurally match B, therefore A|B can match B
> local c = ab or b

> Perhaps it should be special casing where there is a __is metamethod?

I'm going to think out loud here, but my gut reaction is that with nominal typing (given that `__is` only appears in record/interface declarations, where things have names) that shouldn't have been necessary.

Matching nominal against non-nominal records is meant for things like assigning literal tables, which don't have `__is`. If both sides of a comparison are types with `__is`, they should have come from two nominals, and then nominal comparison should have solved that.

Which then leads us to...

> -- ambiguous: A can structurally match B, therefore A|B can match B
> local c = ab or b

My instinct here is that this should have been solved by a higher-priority rule for union types.

The reason there's so much logic around `or` is to try to make things ergonomic for users, trying to go for what the usual intention of an expression is, instead of just doing type propagation and unification, which often leads to surprising error messages. In other words, in theory we could just always translate `x or y` of types `X` and `Y` to `X | Y` and call it a day, but then whenever a programmer makes a mistake in an `or` expression a new union type would pop up.

In this case, it feels natural that the type of `A|B or B` should be `B`, based on their nominal equality alone, regardless of their structure.

mbartlett21:
> > However, they seem to mostly be nominal vs resolved types. (which I have an individual commit on that at https://github.com/mbartlett21/tl/tree/nominal-comparison which I can merge into here if you would like)
> 
> This looks okay, but without fully thinking it through, I'm a bit wary wondering if it slides us further towards structural typing.

I might look into it further. It seemed to be that the types were already resolved on one side, so it was being ambiguous between `{Type}` and `{interface Type}` in the specific case I was looking at.

> ...
>
> In this case, it feels natural that the type of `A|B or B` should be `A|B`, based on their nominal equality alone, regardless of their structure.

That makes sense. That was what my special casing was doing where if there was a union on one side but not the other it preferred the union, even though with structural typing you could get the other.

For structural typing to work there, it would have to have the shared field there with a different type, such as a single item enum. Or have tl special case `__is` metamethod and make sure they are always followed.

hishamhm:
Superseded by #983

PR #983, opened 2025-05-17
update typechecking for `or` expressions, take 2

hishamhm:
Combines #979 and #980, adding a fix to avoid dropping nominal types into their structures too early during inference.

### DIFF
--- a/spec/lang/inference/or_spec.lua
+++ b/spec/lang/inference/or_spec.lua
@@ -56,4 +56,20 @@ describe("inference in 'or' expressions", function()
    ]], {
       { msg = "unused function f" }
    }, {}))
+
+   it("infers `A|B or B` to `A|B`, avoiding ambiguity", util.check_warnings([[
+      local record A where self.field == "a"
+         field: string
+      end
+      local record B where self.field == "b"
+         field: string
+      end
+
+      local ab: A|B
+      local b: B
+
+      local c = ab or b
+   ]], {
+      { msg = "unused variable c: A | B" }
+   }, {}))
 end)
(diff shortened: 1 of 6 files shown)


spec/lang/inference/or_spec.lua
```diff
@@ -12,4 +12,64 @@ describe("inference in 'or' expressions", function()
...
+
+      local function f(t: Type)
+         t = g(t)
+         local _ = t is UnionType and t.types or { t }
+      end
+   ]], {
+      { msg = "unused function f" }
+   }, {}))
+
+   it("infers `A|B or B` to `A|B`, avoiding ambiguity", util.check_warnings([[
```
mbartlett21, line 60: Should there be another test checking that `A or B` errors? (or would it assume they are structurally equal?)

tl.tl
```diff
@@ -13757,22 +13756,42 @@ do
...
-                        local b_is = self:is_a(ub, expected)
-                        if a_is and b_is then
-                           t = self:infer_at(node, expected)
-                        end
-                     end
-                     if not t then
-                        local larger_type = b_ge_a and ub or ua
-                        t = larger_type
-                     end
+                  local is_same = self:same_type(ra, rb)
```
mbartlett21, line 13772: Should this be ua and ub like above or does that not work?

mbartlett21: @hishamhm 
I've tested with using unresolved types and it found another couple of mistakes in the compiler. I did try it, but it then breaks table statements with `or` etc. I'd think just leave it as is for the moment and it can be changed to the unresolved types once we have some better inferencing for `table -> nominal`.

spec/lang/inference/or_spec.lua
mbartlett21: It assumes they are structurally equal. Changing it to look at nominal types makes it error like it should, but then breaks `table literal or Type`.

mbartlett21 (approved):

hishamhm:
Merged! \o/

PR #985, opened 2025-05-17
ci: add action for pushing docs updates

hishamhm:
TODO:
* [x] need to pin actions used to their latest commit hashes
* [x] need to find a way to minimally test this (add a manual trigger option and see what happens...)


hishamhm:
TODO:
* need to pin actions used to their latest commit hashes
* need to find a way to minimally test this (add a manual trigger option and see what happens...)

PR #988, opened 2025-05-19
Fix some tests to actually check properly

mbartlett21:
There was a few ones that weren't actually checking properly.


hishamhm:
Thanks for catching those!

That's what we get for writing tests in Lua instead of Teal :grimacing: 

(But still, I think keeping it in Lua is fine if only to avoid the chicken-and-egg situation — also the reason why the compiler build uses that two-step Makefile instead of relying on Cyan...)

hishamhm (approved):
For commits that affect only the `spec/` subtree, I've been always prefixing them with `tests: `; it makes it easier for me to scan them when looking for features when updating the changelog.

I see it's still marked as draft, so I assume you intend to make more changes -- but feel free to commit those whenever you think they're ready!

PR #991, opened 2025-05-21
Fix scope balancing and symbol scope

wu4:
Previously, `commit_scope_transaction` did not respect the scope balancing pattern that requires calling `end_symbol_list_scope` - this would result in symbol output where the scope counts up a large amount, and improper types would be received as well.

Additionally, `tl.symbols_in_scope` would erroneously pick the outermost representation of a symbol, even if inner representations exist. I threw in a condition to invert that behavior.

Also included are regression tests that should hopefully keep either of these from being a problem in the future.

### DIFF
--- a/spec/api/symbols_in_scope_spec.lua
+++ b/spec/api/symbols_in_scope_spec.lua
@@ -1,0 +1,197 @@
+local tl = require("tl")
+
+local function resolve_type_ref(tr, type_number)
+   local type_info = tr.types[type_number]
+   if type_info and type_info.ref then
+      return resolve_type_ref(tr, type_info.ref)
+   else
+      return type_info
+   end
+end
+
+local DUMMY_FILENAME = "test"
+
+local function assert_scope(tr)
+   local symbols = tr.symbols_by_file[DUMMY_FILENAME]
+
+   local scope_level = 0
+
+   for _, s in ipairs(symbols) do
+      if s[3] == "@{" then
+         scope_level = scope_level + 1
+      elseif s[3] == "@}" then
+         scope_level = scope_level - 1
+         assert(scope_level >= 0, "scope went below 0")
+      end
+   end
+
+   assert(scope_level == 0, "scope is non-zero at end")
+end
+
+local function assert_types(tr, type_locations)
+   for _, location in ipairs(type_locations) do
+      local scoped = tl.symbols_in_scope(tr, location.line, 0, DUMMY_FILENAME)
+      local scoped_type = assert(scoped[location.name] or tr.globals[location.name], "no type data found for " .. location.name .. " at line " .. tostring(location.line))
+      local type_info = resolve_type_ref(tr, scoped_type)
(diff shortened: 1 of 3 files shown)


hishamhm:
> Additionally, tl.symbols_in_scope would erroneously pick the outermost representation of a symbol, even if inner representations exist. I threw in a condition to invert that behavior.

Oh, now that you described it this way, I think this particular behavior might be related to type inference. Because a variable might be declared with a type, but then type inference temporarily "changes its type" in an inner scope. Arguably, both informations are relevant: the variable's true type at definition, and the variable's value currently inferred type at the cursor position. IIRC people have already expected the opposite behavior that the tl API was currently exposing (@pdesaulniers?)

I see that `TypeCollector` [does not make a distinction](https://github.com/teal-language/tl/blob/master/tl.tl#L8243-L8245) if there's a type specialization going on (which at this point would be easy to check). I'm not sure if we want to extend the symbol information with that info, and even if we do, that should probably go in a different PR.

But since you mentioned that teal-language-server was already working around the current behavior (and I think vscode-teal as well?) and implementating their own behavior more in line with this PR, then I think this should be the way to go indeed!

spec/lang/parser/blocks_spec.lua
```diff
@@ -0,0 +1,197 @@
+local tl = require("tl")
```
hishamhm, line 1: could you move this file instead to `spec/api/symbols_in_scope_spec.lua`?

hishamhm:
You can squash together the last two commits when moving the file, and give the commit a `tests:` prefix -- thanks!

wu4:
> Oh, now that you described it this way, I think this particular behavior might be related to type inference. Because a variable might be declared with a type, but then type inference temporarily "changes its type" in an inner scope. Arguably, both informations are relevant: the variable's true type at definition, and the variable's value currently inferred type at the cursor position. IIRC people have already expected the opposite behavior that the tl API was currently exposing (@pdesaulniers?)
> [...]
> But since you mentioned that teal-language-server was already working around the current behavior (and I think vscode-teal as well?) and implementating their own behavior more in line with this PR, then I think this should be the way to go indeed!

* The pre-commit code would fail some of those tests - any variables that are shadowed will be seen as the "original" type instead. Most notably `self` won't be typed correctly in, for example, nested record methods.
* Specifically, the name of the function `symbols_in_scope` implies to me that the symbols would point to the types relevant to them within that scope.
* In fact, `teal-language-server`'s symbol-type resolution actually expected it to work that way, and it fails as a result. There is a [code block](https://github.com/teal-language/teal-language-server/blob/main/src/teal_language_server/document.tl#L451) dedicated to specifically resolving `self` through string parsing, but it leaves the types of other symbols incorrect to the scope.

To summarize, I believe that if someone needs access to all of the symbols available to the file, they have access to `TypeReport.symbols_by_file` as well as `TypeReport.globals`. I think that `tl.symbols_in_scope` should cover the niche of viewing the available types from the inside of a scope out. Given the name, culling unavailable types to a scope makes sense, but providing the closest-to-global type representation of a symbol rather than respecting the scope is unintuitive.

hishamhm:
Makes sense! Merged!

PR #996, opened 2025-05-26
Check for missing fields when subtyping records

mbartlett21:
This does also break a fair bit of stuff (interestingly no tests)

The compiler was only filling out the `i` field on tokens on the `$EOF$` token. Let me know if you want it to not fill it out at all or whether you are happy with exposing that field everywhere.

### DIFF
--- a/tl.tl
+++ b/tl.tl
@@ -670,6 +670,7 @@ local record tl
    record Token
       x: integer
       y: integer
+      i: integer
       tk: string
       kind: TokenKind
    end
@@ -1132,6 +1133,7 @@ do
          tokens[nt] = {
             x = tx,
             y = ty,
+            i = ti,
             tk = tk,
             kind = kind,
          }
@@ -1144,8 +1146,9 @@ do
          tokens[nt] = {
             x = tx,
             y = ty,
+            i = ti,
             tk = tk,
-            kind = keywords[tk] and "keyword" or "identifier"
+            kind = keywords[tk] and "keyword" or "identifier",
          }
          in_token = false
       end
@@ -1156,8 +1159,9 @@ do
          tokens[nt] = {
             x = tx,
             y = ty,
+            i = ti,
             tk = tk,
-            kind = kind
+            kind = kind,
          }
          in_token = false
       end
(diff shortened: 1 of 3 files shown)


hishamhm:
I believe some of these lenient behaviors on things like matching literal tables date back from the earliest versions of the compiler when the typing rules weren't as well-defined.

But looking at the changes to tl itself, it seems that it has only caught actual mistakes. I want to give this a run on the LuaRocks codebase later and see how it fares there. 

I think we want this, the question is whether this is a "Teal 2024 bugfix" or "Teal 2025 change".

tl.tl
```diff
@@ -670,6 +670,7 @@ local record tl
    record Token
       x: integer
       y: integer
+      i: integer
```
hishamhm, line 673: I don't think this value is actually used anywhere, right? I think the `i` in `$EOF$` was a leftover.

mbartlett21: I'd think it would be nice to expose it to give the original indices into the source, rather than having to reconstruct it from x and y. Perhaps renamed to i_start or something.

hishamhm: In the Zig compiler, they actually removed the `x` and `y` entries and only kept the `i` entries, and they derive the line/column info on demand, only when displaying error message (the assumption being that correct code is compiled much more often, so performance should be optimized towards that case: https://youtu.be/IroPQ150F6c?t=2002 ). I briefly considered it, but I think in general it's a lot more useful for us at this stage to have that info easily readable for debugging compared to the minute performance benefit of one fewer integer in a table.

> rather than having to reconstruct it from x and y

Is anybody currently having to reconstruct it? (Not a rhethorical question, I'm just not aware if any other tooling currently needs it.)

hishamhm: I'd say let's just leave it out until some time we actually need it, then!

mbartlett21:
@hishamhm 
Did you want this rebased on your split branch or do I wait till it gets merged?

hishamhm:
@mbartlett21 I think it's fine to keep it all together! It's a small tweak to the Makefile.

Also, looks like my locally generated tl.lua included some leftover files in it. I'll need to clean up my local environment and make the script a bit smarter!

PR #1005, opened 2025-06-06
Pass comments to AST nodes during parsing

upedd:
Part of tealdoc project. Passes comments to AST nodes of all declarations, record/interface fields and enum values.

### DIFF
--- a/spec/lang/parser/store_comments_spec.lua
+++ b/spec/lang/parser/store_comments_spec.lua
@@ -1,0 +1,461 @@
+local tl = require("tl")
+local util = require("spec.util")
+
+describe("store comments in syntax tree", function() 
+    it("comments before implicit global function", function() 
+        local result = tl.process_string([[
+            -- this is a comment
+            function --[==[ignore me]==] foo() end
+        ]])
+        assert.same({}, result.syntax_errors)
+        assert.same(1, #result.ast)
+        assert.same("statements", result.ast.kind)
+        assert.same("global_function", result.ast[1].kind)
+        assert.same("-- this is a comment", result.ast[1].comments[1].text)
+    end)
+    it("comments before local function", function() 
+        local result = tl.process_string([[
+            -- this is a comment
+            local function --[==[ignore me]==] foo() end
+        ]])
+        assert.same({}, result.syntax_errors)
+        assert.same(1, #result.ast)
+        assert.same("statements", result.ast.kind)
+        assert.same("local_function", result.ast[1].kind)
+        assert.same("-- this is a comment", result.ast[1].comments[1].text)
+    end)
+    it("comments before local variable declaration", function() 
+        local result = tl.process_string([[
+            -- this is a comment
+            local --[==[ignore me]==] x = 42
+        ]])
+        assert.same({}, result.syntax_errors)
+        assert.same(1, #result.ast)
+        assert.same("statements", result.ast.kind)
+        assert.same("local_declaration", result.ast[1].kind)
+        assert.same("-- this is a comment", result.ast[1].comments[1].text)
+    end)
+    it("comments before local type declaration", function() 
+        local result = tl.process_string([[
+            -- this is a comment
+            local --[==[ignore me]==] type Foo = number
+        ]])
+        assert.same({}, result.syntax_errors)
+        assert.same(1, #result.ast)
+        assert.same("statements", result.ast.kind)
+        assert.same("local_type", result.ast[1].kind)
+        assert.same("-- this is a comment", result.ast[1].comments[1].text)
+    end)
+    it("comments before local macroexp", function() 
+        local result = tl.process_string([[
+            -- this is a comment
+            local --[==[ignore me]==] macroexp foo(): number
+                return 2
+            end
+        ]])
+        assert.same({}, result.syntax_errors)
+        assert.same(1, #result.ast)
+        assert.same("statements", result.ast.kind)
+        assert.same("local_macroexp", result.ast[1].kind)
+        assert.same("-- this is a comment", result.ast[1].comments[1].text)
+    end)
+    it("comments before local record declaration", function() 
+        local result = tl.process_string([[
+            -- this is a comment
+            local --[==[ignore me]==] record Foo
+                x: number
+            end
+        ]])
+        assert.same({}, result.syntax_errors)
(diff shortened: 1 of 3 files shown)


tl.tl
```diff
@@ -1883,8 +1883,10 @@ local interface RecordLikeType
    interfaces_expanded: boolean
    fields: {string: Type}
    field_order: {string}
+   field_comments: {string: {{Comment}}}
    meta_fields: {string: Type}
    meta_field_order: {string}
+   meta_field_comments: {string: {{Comment}}}
```
hishamhm, line 1889: @upedd It's unfortunate that we have the storage of comments mixed across Node and Type, but I understand why you did it this way.

There is a backstory as to why the parser mixes the creation of Node and Type objects — back in the beginning, it felt like a small optimization. But then when I tried to change everything into Node for a better separation of concerns, it became apparent that the checker now had a dependency on this behavior: this is because the type-checker runs as a single pass of the Node tree, and for some situations it becomes suspiciously convenient that the Type objects are already there during the "first pass" (that is, the processing of Type objects as part of the parser functions like a "hidden zero-th pass").

This is eventually fixable, but not something I want us to do right away because changing that has other consequences to the later checking stage (there are other reasons why I'm considering turning the type-checker into a more (nowadays) common multi-pass process).

So we'll keep it the way you did it for now — this was more for sharing and "documenting" the context behind it.

(cc @Frityet since this is also important for you to know)

tl.tl
```diff
@@ -2010,6 +2012,7 @@ end
 
 -- Intersection types, currently restricted to polymorphic functions
 -- defined inside records, representing polymorphic Lua APIs.
+-- types order should match the order of declarations in the record or interface.
```
hishamhm, line 2015: Perhaps this should be a comment on the types field itself?

upedd:
Thank you for your thorough review, @hishamhm. I’ve fixed all the issues you pointed out.

Also noticed and fixed some flawed logic in the test code.

spec/lang/parser/store_comments_spec.lua
```diff
@@ -0,0 +1,498 @@
...
+        assert.same("statements", result.ast.kind)
+        assert.same("local_type", result.ast[1].kind)
+        assert.same("newtype", result.ast[1].value.kind)
+        local record_def = result.ast[1].value.newtype.def
+        assert.same("record", record_def.typename)
+        local expected_comments = {
+            ["f"] = {nil, "--- it can be a boolean too", nil}
+        }
+        for field_name, _ in pairs(record_def.fields) do
+            for i = 1, #record_def.field_comments[field_name] do
```
hishamhm, line 489: Lua footgun here: the `#` operator is not "stable" if there may be `nil` entries in your array part (if you look up the definition of `#` in the Lua Reference Manual you will see how subtle it is; but in short, if you have `t = {"a", nil, nil, "d", "e", nil, "g"}`, then `#t` may return either 1, 5 or 7 — this is because Lua does not store the "array size" in the table, but performs a binary search to find the last element — it is time-efficient enough and much more space-efficient, all tables considered).

You can fix this here by simply using the `fields` count, which never has `nil` entries:

```suggestion
            for i = 1, #record_def.fields[field_name] do
```

spec/lang/parser/store_comments_spec.lua
```diff
@@ -0,0 +1,498 @@
...
+        assert.same("statements", result.ast.kind)
+        assert.same("local_type", result.ast[1].kind)
+        assert.same("newtype", result.ast[1].value.kind)
+        local record_def = result.ast[1].value.newtype.def
+        assert.same("record", record_def.typename)
+        local expected_comments = {
+            ["__call"] = "-- this is a comment",
+            ["__add"] = "-- another comment",
+        }
+        for field_name, _ in pairs(record_def.fields) do
```
hishamhm, line 277: This is iterating over `fields` and not `meta_fields`

tl.tl
```diff
@@ -3853,9 +3861,23 @@ local function parse_return(ps: ParseState, i: integer): integer, Node
...
+local function store_field_in_record(ps: ParseState, i: integer, field_name: string, newt: Type, def: RecordLikeType, comments?: {Comment}, meta?: MetaMode): boolean
+   local field_order, fields, field_comments: {string}, {string:Type}, {string: {{Comment}}}
+   if meta then
+      field_order, fields, field_comments = def.meta_field_order, def.meta_fields, def.meta_field_comments
+   else
+      field_order, fields, field_comments = def.field_order, def.fields, def.field_comments
+   end
+
+   if comments and not field_comments then 
+      field_comments = {}
```
hishamhm, line 3873: this new table isn't being attached back into `def`.

I can see that this is working because you're still always creating one below for `def.field_comments` (line 4097), but I don't see an equivalent for `def.meta_field_comments`, which made me wonder how the tests were passing.

But upon a closer look at the tests, I see that the loop in the test was iterating over the wrong table resulting in an empty loop (see the other comment above).

tl.tl
```diff
@@ -4053,6 +4094,7 @@ end
 parse_record_body = function(ps: ParseState, i: integer, def: RecordLikeType): integer, boolean
    def.fields = {}
    def.field_order = {}
+   def.field_comments = {}
```
hishamhm, line 4097: by initializing it in `store_field_in_record`, this can go away

tl.tl
```diff
@@ -3914,17 +3952,20 @@ local function parse_nested_type(ps: ParseState, i: integer, def: RecordLikeType
...
    nt.newtype = new_typedecl(ps, istart, ndef)
 
-   store_field_in_record(ps, iv, v.tk, nt.newtype, def.fields, def.field_order)
+   store_field_in_record(ps, iv, v.tk, nt.newtype, def, ps.tokens[istart].comments)
    return i
 end
 
 parse_enum_body = function(ps: ParseState, i: integer, def: EnumType): integer, boolean
    def.enumset = {}
+   def.value_comments = {}
```
hishamhm, line 3961: can we make this one on-demand as well, only when `item.comments` exists below?

spec/lang/parser/store_comments_spec.lua
upedd: I don't think that is really needed, as the `store_field_in_record` never passes nil entries to this array

hishamhm (approved):

PR #1008, opened 2025-06-18
Set the main build to target Lua 5.1

mbartlett21:
This makes it convert `//` ops into the appropriate `math.floor(x/y)` to be compatible with Lua 5.1


hishamhm:
@mbartlett21 I think it's fine to keep it in the command line. I'm wondering if we should use an environment variable that we set on CI?

hishamhm (approved):

PR #1023, opened 2025-08-14
fix: TypeChecker:expand_interfaces should copy array elements, not the array

### LINKED ISSUE #1022: Odd array interface inheritance error
I expect the following code to compile fine

```teal
local interface A is {string}
end

local interface B is A
end

local a: A = { "foo", "bar" }
local b: B = { "foo", "bar" }
print(a[1])
print(b[1])
```

But it produces some errors:

```console
$ tl check repro.tl
2 errors:
repro.tl:8:16: in local declaration: b: expected an array: at index 1: got string "foo", expected {string}
repro.tl:8:23: in local declaration: b: expected an array: at index 2: got string "bar", expected {string}
----------------------------------------
2 errors
```

The compiler seems to think that the array component of `B` is `{string}` rather than `string`. More worryingly, the code is accepted by the compiler when `b` is initialized to `{ { "foo", "bar" } }`.

After a very brief poking around the error is fixed by adding this simple change to `TypeChecker:expand_interfaces`

```diff
index e848060..f021d39 100644
--- a/tl.tl
+++ b/tl.tl
@@ -14461,7 +14461,7 @@ do
                end
             else
                if not t.elements then
-                  t.elements = iface
+                  t.elements = iface.elements
                else
                   if not self:same_type(iface.elements, t.elements) then
                      self.errs:add(t, "incompatible array interfaces")
```

It doesn't seem to break any tests, so hopefully that's it? Happy to add a regression test in a pr.

euclidianAce:
closes #1022

### DIFF
--- a/spec/lang/subtyping/interface_spec.lua
+++ b/spec/lang/subtyping/interface_spec.lua
@@ -247,4 +247,13 @@ describe("subtyping of interfaces:", function()
       { y = 13, msg = "I is not a R" },
       { y = 14, msg = "I is not a R" },
    }))
+
+   it("array inheritance works (regression test for #1022)", util.check([[
+      local interface A is {string}
+      end
+      local interface B is A
+      end
+      local _a: A = { "a", "b" }
+      local _b: B = { "c" }
+   ]]))
 end)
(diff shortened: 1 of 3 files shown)


hishamhm:
@euclidianAce thank you!

PR #1028, opened 2025-09-01
GSoC 2025 - Macro support for teal

Frityet:
This PR implements macro support for Teal.

[Project Proposal](https://summerofcode.withgoogle.com/programs/2025/projects/kiO8b1Ql)

### DIFF
--- a/t.tl
+++ b/t.tl
@@ -1,0 +1,330 @@
+local macro match!(value: Expression, cases: Expression): Expression
+   local blk = require("teal.block")
+   local BI = blk.BLOCK_INDEXES
+
+   local function or_chain(exprs: {any})
+      if #exprs == 0 then return nil end
+      local acc = exprs[1]
+      for i = 2, #exprs do
+         local e = exprs[i]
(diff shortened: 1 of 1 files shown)


spec/lang/macro/invocation_spec.lua
```diff
@@ -0,0 +1,597 @@
...
+               print(x)
+            end
+         , 'ok')
+      ]]
+      local ast, errs = tl.parse(code)
+      assert.same({}, errs)
+      local out, err = lua_gen.generate(ast, '5.4')
+      assert.is_nil(err)
+      out = out:gsub('%s+', ' ')
+      assert.match('for _, x in ipairs%(%s*t%s*%) do%s*print%(x%)%s*end', out)
```
hishamhm, line 236: We generally avoid complex pattern-matching in test cases like these, which make maintenance and future debugging harder. Instead, just to an exact string comparison. Also, we have util functions for testing code generation. Please use those rather than using the API directly.

PR #1051, opened 2025-11-24
GSoC 2025 - Macro support for teal

Frityet:
This PR implements macro support for Teal.

[Project Proposal](https://summerofcode.withgoogle.com/programs/2025/projects/kiO8b1Ql)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ BUSTED = busted --suppress-pending
 PRECOMPILED = teal/precompiled/default_env.lua
 SOURCES = teal/debug.tl teal/attributes.tl teal/errors.tl teal/lexer.tl \
 	teal/reader.tl teal/block.tl \
-	teal/util.tl teal/types.tl teal/facts.tl teal/parser.tl teal/traversal.tl \
+	teal/util.tl teal/types.tl teal/facts.tl teal/ast.tl teal/parser.tl teal/traversal.tl \
 	teal/variables.tl teal/type_reporter.tl \
 	teal/macroexps.tl teal/macro_eval.tl teal/metamethods.tl \
 	teal/type_errors.tl teal/environment.tl \


Frityet:
> seems like nodes made with `block()` aren't properly showing up in generated lua? investigating...
> 
> ```lua
> local macro test!(): Expression
>     local s = block("string")
>     s.conststr = "hello"
>     return s
> end
> 
> local x = test!()
> ```

nevermind, this is intentional, its cause I didn't put in `tk`... @hishamhm should we make it so if its a block that depends on `tk`, such as `string` or `integer` that it should error?

teal/check/require_file.lua
```diff
@@ -1,4 +1,4 @@
-local _tl_compat; if (tonumber((_VERSION or ''):match('[%d.]*$')) or 0) < 5.3 then local p, m = pcall(require, 'compat53.module'); if p then _tl_compat = m end end; local io = _tl_compat and _tl_compat.io or io; local os = _tl_compat and _tl_compat.os or os; local package = _tl_compat and _tl_compat.package or package; local string = _tl_compat and _tl_compat.string or string; local table = _tl_compat and _tl_compat.table or table; local input = require("teal.input")
+local _tl_compat; if (tonumber((_VERSION or ''):match('[%d.]*$')) or 0) < 5.3 then local p, m = true, require('compat53.module'); if p then _tl_compat = m end end; local io = _tl_compat and _tl_compat.io or io; local os = _tl_compat and _tl_compat.os or os; local package = _tl_compat and _tl_compat.package or package; local string = _tl_compat and _tl_compat.string or string; local table = _tl_compat and _tl_compat.table or table; local input = require("teal.input")
```
hishamhm: @Frityet all of these generated .lua files contain a change from `pcall(require, ...)` to `true, require(...)`. Did you generate with `--gen-compat=required`? Could you change it to `--gen-compat=optional`? This should reduce the PR diff a lot!

teal/macro_eval.tl
```diff
@@ -0,0 +1,389 @@
...
+   --todo: restrict
+   local FORBIDDEN_LIBS: {string:boolean} = {
+      os = true,
+      io = true,
+      debug = true,
+      package = true,
+      load = true,
+      loadfile = true,
+      dofile = true,
+      -- require = true,
```
hishamhm: What's the current story with `require` in macros? Are we letting it require code unrestricted?

teal/macro_eval.tl
```diff
@@ -0,0 +1,389 @@
...
+   local lua_generator = require("teal.gen.lua_generator")
+   local single: Block = { kind = "statements", y = mb.y, x = mb.x, tk = mb.tk, yend = mb.yend, xend = mb.xend }
+   single[1] = mb
+   local mast, perrs = parser.parse(single, filename, read_lang)
+   if #perrs > 0 then
+      for _, e in ipairs(perrs) do table.insert(errs, e) end
+      return
+   end
+
+   local code = lua_generator.generate(mast as ParserNode, "5.1", lua_generator.fast_opts)
```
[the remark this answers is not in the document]
hishamhm: > I was thinking maybe we should just do it as whatever lua version tl is running? 

No, we did this in LuaRocks and it does get really mixed up. Teal has a `--gen-target`. We either do lowest-common-denominator, or we reuse whatever the user is using as a `--gen-target`. But if we follow the gen_target, then we'll start having Teal code with macros that only works on a single Lua version. We should avoid that if we can, so I think lowest-common-denominator for the macro bodies should suffice for now.

teal/init.tl
```diff
@@ -304,7 +305,17 @@ end
...
+   local ast, parse_errs, parse_required = parser.parse(block_ast, self.filename)
+   for _, e in ipairs(parse_errs) do
+      table.insert(errs, e)
+   end
+   if parse_required then
+      required_modules = required_modules or {}
+      for _, m in ipairs(parse_required) do
+         table.insert(required_modules, m)
+      end
+   end
```
hishamhm: Same here

teal/input.tl
```diff
@@ -15,7 +16,15 @@ function input.check(env: Env, filename: string, code: string): Result
...
-   local program, syntax_errors = parser.parse(code, filename)
+   local block_ast, read_errors = reader.read(code, filename)
+   local program, parse_errors = parser.parse(block_ast, filename)
+   local syntax_errors = {}
+   for _, e in ipairs(read_errors) do
+      table.insert(syntax_errors, e)
+   end
+   for _, e in ipairs(parse_errors) do
+      table.insert(syntax_errors, e)
+   end
```
hishamhm: Same here

teal/reader.tl
```diff
@@ -0,0 +1,2815 @@
...
+   if optional_paren then
+      i = verify_tk(ps, i, ")")
+   end
+
+   return i, t
+end
+
+local function read_function_args_rets_body(ps: ReadState, i: integer, node: Block): integer, Block
+   local istart = i - 1
+   --TODO: integrate types
```
hishamhm: I guess the comments in this function are stale and can be removed?

Frityet:
> Macro support sounds cool, is there documentation about them somewhere already?
> 
> 
> 
> The tests don't cover much when it comes to that :(

See #1085

PR #1057, opened 2025-12-19
CI: cache luarocks folder

### LINKED ISSUE #858: CI: we should cache Busted and its dependencies
We shouldn't need to build Busted and friends every time on Github Actions. This is especially slow on Windows (1min15 vs 25s in the Linux runners)

santimorph:
Should fix #858.

I went for the whole folder instead of just Busted and it's dependencies because doing just Busted would've left me trying to reference multiple specific paths per platform and that felt way too messy. Hope that's ok.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,7 @@ jobs:
     - uses: luarocks/gh-actions-luarocks@master
 
     - name: Restore cached LuaRocks packages
+      id: cache
       uses: actions/cache/restore@v4
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -37,3 +44,10 @@ jobs:
         luarocks lint tl-dev-1.rockspec
         make selfbuild
         luarocks test
+
```
alerque: This is quite a bit of hoopla and repeated code you don't need to do.

If you just use `actions/cache@v4` instead of manually using the save and restore functions separately then you only need to have one `uses:` block (at the point where the restore would be) and a hook takes care of the save automatically. See here for an example in use:

https://github.com/alerque/polytype/blob/master/.github/workflows/deploy.yml#L27-L31

PR #1063, opened 2026-01-08
Fix __call self

### LINKED ISSUE #1060: Interface __call returning `self` resolves to wrong type when passed to method
When an interface with `__call` returning `self` is passed to a method, the return type incorrectly resolves to the record containing the method.

### Environment

- tl version: 0.24.7+dev

### Minimal reproduction

```teal
local record Foo
end

local interface ICallable
    metamethod __call: function(self, ...: any): self
end

function Foo:bar(a: ICallable): integer
    return 1
end

local x: ICallable
local foo: Foo = {}

foo:bar(x(1))  -- ERROR
```

Expected behavior:

x(1) should return ICallable, so foo:bar(x(1)) should type check successfully.

Actual behavior:

```
1 error:
argument 1: got Foo, expected interface ICallable
```

The self return type of __call is incorrectly resolved to Foo.

Workaround:

```teal
-- Explicit cast
foo:bar((x as function(integer): ICallable)(1))
```

> Note: This may be related to #931 (generic type resolution with interfaces).

mtdowling:
Closes #1060


hishamhm:
Thank you both for looking into this! I haven't looked closely at it yet, but from reading the PR and patch, I wonder if the root cause is that we're missing a call to `Context:resolve_self`, or if we're calling it at the wrong place...

PR #1069, opened 2026-01-15
ci: streamline cache flow

santimorph:
Going by https://github.com/teal-language/tl/pull/1057#discussion_r2635650743 it's better to just run the base cache action on the restore spot and it saves automatically.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,9 +28,9 @@ jobs:
 
     - uses: luarocks/gh-actions-luarocks@master
 
-    - name: Restore cached LuaRocks packages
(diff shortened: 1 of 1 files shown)


hishamhm:
@santimorph Thank you! Merged!

Aire-One:
Reading the doc at https://github.com/actions/cache?tab=readme-ov-file#restoring-and-saving-cache-using-a-single-action. Shouldn't the installation step have a `if: steps.cache-XXX.outputs.cache-hit != 'true'` clause to prevent it from running when the cache is hit? 

Btw, should lua also be cached?

PR #1074, opened 2026-01-21
Adjust playground building

mbartlett21:
Lets try again...


mbartlett21:
I'll commit this to main and then check again for it

PR #1083, opened 2026-02-01
Fix resolution of global type aliases

### LINKED ISSUE #998: "attempt to index a nil value (field 'declared_at')" when parsing global type alias
`var.aliasing.declared_at` can sometimes be nil in `check_var_usage` which causes an exception to be thrown when parsing teal code with `tl check`

Code to reproduce:
```
global record TestRecord
  new: function(): TestRecord
end

global type TestRecordAlias = TestRecord

local foo = TestRecordAlias.new()
```

Stack Trace:
```
[string "/home/runner/work/tl/tl/_binary/src/tl.lua"]:6661: attempt to index a nil value (field 'declared_at') stack traceback:
[string "/home/runner/work/tl/tl/_binary/src/tl.lua"]:6661: in upvalue 'check_var_usage'
[string "/home/runner/work/tl/tl/_binary/src/tl.lua"]:6674: in method 'check_var_usage'
[string "/home/runner/work/tl/tl/_binary/src/tl.lua"]:14723: in function 'tl.check'
[string "/home/runner/work/tl/tl/_binary/src/tl.lua"]:14861: in function 'tl.check_string'
```
Image

hishamhm:
Fixes #998.

### DIFF
--- a/spec/lang/code_gen/global_type_spec.lua
+++ b/spec/lang/code_gen/global_type_spec.lua
@@ -100,4 +100,22 @@ describe("global type code generation", function()
 
       print(c2:use_it())
    ]]))
+
+   it("does not elide global record type used in alias (#998)", util.gen([[
+      global record TestRecord
+         new: function(): TestRecord
+      end
+
+      global type TestRecordAlias = TestRecord
+
+      local foo = TestRecordAlias.new()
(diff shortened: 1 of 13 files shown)


luisvsm:
Thank you! <3

PR #1091, opened 2026-02-10
Propagate contextual expected function types to function literals

Nikita-baghela07:
This change propagates the contextual expected type into function literal (lambda) nodes during type checking.

When a function expression appears in a context that already defines a function type (e.g., variable assignment, argument position, or return position), the expected type is now made available to the function literal.

This PR does not change any typing behavior yet. It only ensures that the expected function type is accessible to the lambda node so that future work can implement contextual parameter type inference.

### DIFF
--- a/spec/lang/inference/expected_function_types_spec.lua
+++ b/spec/lang/inference/expected_function_types_spec.lua
@@ -1,7 +1,8 @@
 local util = require("spec.util")
 
-describe("expected function type propagation", function()
+describe("function definition type propagation", function()
 
+   -- Argument type propagation
    it("propagates expected type in variable assignment", util.check([[
       local f: function(integer): string = function(x)
          local added: integer = x + 1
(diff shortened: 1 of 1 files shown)


spec/lang/inference/expected_function_types_spec.lua
```diff
@@ -0,0 +1,37 @@
+local util = require("spec.util")
+
+describe("expected function type propagation", function()
+
+   it("propagates expected type in variable assignment", util.check([[
+      local f: function(integer): string = function(x)
+         return tostring(x)
+      end
```
mbartlett21: This doesn't test that x is a number-like type.

It should be something like below:
```teal
local f: function(integer): string = function(x)
   local added: integer = x + 1
   return tostring(added)
end
```

And this applies to all the examples below with `tostring(any)`

spec/lang/inference/expected_function_types_spec.lua
```diff
@@ -0,0 +1,37 @@
+local util = require("spec.util")
```
mbartlett21: Are you happy to name this file something like `function_definition_spec` or similar to match with the rest of the test files?

Issue #1029, opened 2025-09-10
Feature Request: Allow records to also be tuples


JoshieGemFinder:
Allowing records to be tuples (if they aren't already arrays) would allow for much more flexibility in the data they can represent, especially when interacting with vanilla lua modules where that situation isn't uncommon.

The syntax would be simple, re-using the `is` syntax for arrays but with a tuple typing instead:
```lua
local type MyTupleType = {number, number, string}

-- maybe this would be representative of a table that's returned by a function from an external vanilla lua module
local record MyRecordType is MyTupleType
    foo: number

    -- macros to make reading the values out of the tuple easier
    get_x: function(self: MyRecordType): number = macroexp(self: MyRecordType): number return self[1] end
    get_y: function(self: MyRecordType): number = macroexp(self: MyRecordType): number return self[2] end
    get_name: function(self: MyRecordType): string = macroexp(self: MyRecordType): string return self[3] end
end

-- some external function that returns values in the form {[1]: number, [2]: number, [3]: number, foo: number}
local some_external_function = require("some/library.lua").some_function as (function(num: number): MyRecordType)

local some_result: MyRecordType = some_external_function(42)

print(some_result:get_x()) -- transpiles to print(some_result[1])
```

I won't define the behaviour for if a record is attempted to be declared both an array and a tuple, but I think it should probably fail to compile like it does currently.

PR #1101, opened 2026-03-06
Allow interfaces and records to have tuple keys

JoshieGemFinder:
I have implemented the functionality for recordlike types to have tupletable parent interfaces (#1029).

The basic syntax is very similar to how it is to define an interface to have an array parent interface:
```lua
-------- Basic example + syntax --------

local interface MyTupleInterface is {number, number, string}
	description: string

	get_x: function(self): number = macroexp(self: MyTupleInterface): number return self[1] end
	get_y: function(self): number = macroexp(self: MyTupleInterface): number return self[2] end
	get_name: function(self): string = macroexp(self: MyTupleInterface): string return self[3] end
end

local tuple_object: MyTupleInterface = {
	[1] = 10,
	[2] = 20,
	[3] = "Example Name",
	description = "Example Description"
}

print(tuple_object:get_x())    -- transpiles to print(tuple_object[1])
print(tuple_object:get_y())    -- transpiles to print(tuple_object[2])
print(tuple_object:get_name()) -- transpiles to print(tuple_object[3])

local obj_name: string = tuple_object[3] -- number indexing is also valid
local obj_description: string = tuple_object.description -- does not interfere with regular field typings
print(obj_name, obj_description)

-- Indexes are limited to the range of the tuple
print(tuple_object[0]) -- error: "index 0 out of range for tuple MyTupleInterface"
print(tuple_object[4]) -- error: "index 4 out of range for tuple MyTupleInterface"
```

If an recordlike type implements parents which are both tuples then it will attempt to combine the tuples, or error if the tuple types are incompatible:
```lua
-------- Compatibility with other tuple interfaces --------

local interface MyCompatibleTupleInterface is {number, number, string, boolean}
	-- This interface is compatible, because {number, number, string} and 
	--     {number, number, string, boolean} share the same starting elements
end

local record MyCompatibleRecord is MyTupleInterface, MyCompatibleTupleInterface
	get_flag: function(self): number = macroexp(self: MyCompatibleRecord): number return self[4] end
end

local interface MyIncompatibleTupleInterface is {number, string, number}
	-- This interface is incompatible, because {number, number, string} and 
	--     {number, string, number} do not share the same elements
end

-- error: "incompatible tuple interfaces"
local record MyIncompatibleRecord is MyTupleInterface, MyIncompatibleTupleInterface

end
```

I currently have opted for the most lenient behaviour when a recordlike type attempts to implement both an array superinterface and a tuple superinterface. Currently, a warning is generated and the tuple types take precedence over the array types where applicable.  
I believe an improvement could be erroring by default, but letting the user manually override the error in some way; however, I wanted to avoid any unnecessarily large code changes in this pull request.

```lua
-------- Compatibility with array interfaces --------

local interface MyArrayInterface is {boolean}

end

-- warn: "inherits from incompatible overlapping array {boolean} and tuple {number, number, string}"
local record MyTupleArrayRecord is MyTupleInterface, MyArrayInterface

end

-- All tuple fields remain valid and keep their type
local tuple_array_object: MyTupleArrayRecord = {
	[1] = 10,
	[2] = 20,
	[3] = "Example Name",
	description = "Example Description"
}

-- These also all work the same
print(tuple_object:get_x())    -- transpiles to print(tuple_object[1])
print(tuple_object:get_y())    -- transpiles to print(tuple_object[2])
print(tuple_object:get_name()) -- transpiles to print(tuple_object[3])

-- But now everything before and after tuple indices is the array's type
tuple_array_object[4] = true -- before the array interface, this would have created an index error
print(tuple_array_object[4]) -- prints "true"

global function test(n: integer)
	-- Variable indices have their union extended to include the array's type as well, just as if it were a tuple element
	-- x is a boolean | number | string
	local x = tuple_array_object[n];
	print(x);
end
```

A couple pieces of my code could probably be made cleaner or more readable, but I had to leave them as-is to avoid significantly restructuring code unrelated to the tuple table interface addition (it would feel rude to forcefully inject my personal code tastes into places it isn't needed). I will leave the decision on that up to the maintainers who will actually be interacting with it.

### DIFF
--- a/spec/lang/subtyping/interface_spec.lua
+++ b/spec/lang/subtyping/interface_spec.lua
@@ -14,6 +14,20 @@ describe("subtyping of interfaces:", function()
       local r: MyRecord = {}
       print(#r)
    ]]))
+   
+   it("record inherits interface tuple definition", util.check([[
+      local interface MyInterface
+         is {integer, string}
+         x: integer
+      end
+
+      local record MyRecord
+         is MyInterface
+      end
+
+      local r: MyRecord = {1, "abc"}
+      print(r[1], r[2])
+   ]]))
 
    it("record <: interface", util.check([[
       local interface MyInterface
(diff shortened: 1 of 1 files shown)


teal/check/context.tl
```diff
@@ -1544,14 +1557,59 @@ do
...
+            end
+         end
+
+         -- Send a warning that this type has overlapping tuple keys and array keys
+         if incompatible_tuple_array then
+            local tuple_types = {}
+            for _, element_type in ipairs(t.types) do
+               table.insert(tuple_types, show_type(element_type))
+            end
+            self.errs:add_warning("inheritance", t, "inherits from incompatible overlapping array {%s} and tuple {%s}", show_type(t.elements), table.concat(tuple_types, ", "))
```
hishamhm: I think this can be a hard error and not a warning. Since this is new syntax, we can opt to be stricter.

JoshieGemFinder: I have updated the code to hard error on incompatible array and tuple typings, and send a warning otherwise

JoshieGemFinder:
As requested, I have updated the code to hard error on incompatible array/tuple typings.

Here is the updated example code:
```lua

-------- Compatibility with array interfaces --------

local interface MySimpleTupleInterface is {string, string}
	-- Tuple interfaces can be compatible with array interfaces only if all types are equal
end

------- Simple compatibility -------

local interface MyCompatibleArrayInterface is {string}
	-- An array of strings is compatible with {string, string}
end

local interface MyIncompatibleArrayInterface is {boolean}
	-- An array of booleans is not compatible with {string, string}
end

-- warn: "inherits overlapping array {string} and tuple {string, string}"
local record MyCompatibleTupleArrayRecord is MySimpleTupleInterface, MyCompatibleArrayInterface

end

-- error: "inherits incompatible array {boolean} and tuple {string, string}"
local record MyIncompatibleTupleArrayRecord is MySimpleTupleInterface, MyIncompatibleArrayInterface

end

------- Compatibility with union type arrays -------

local interface MyUnionArrayInterface is {boolean | string}
	-- An array of {boolean | string} is compatible with {string, string}
end

-- warn: "inherits overlapping array {boolean | string} and tuple {string, string}"
local record MyUnionTupleArrayRecord is MySimpleTupleInterface, MyUnionArrayInterface

end

-- The typing of tuple fields overrides the typing of the array where applicable
local tuple_array_object: MyUnionTupleArrayRecord = {
	[1] = "A",
	[2] = "B"
}

-- error "in assignment: got boolean, expected string"
tuple_array_object[2] = true
-- valid assignments because the array interface accepts both booleans and strings
tuple_array_object[5] = false
tuple_array_object[6] = "G"

-- n1 has a (string) typing instead of (boolean | string) because it's within tuple bounds
local n1 = tuple_array_object[1]
-- n3 has a (boolean | string) typing because [3] is only an array index
local n3 = tuple_array_object[3]

```

JoshieGemFinder:
I believe all of the test specifications should now be in place

hishamhm:
> is this still being considered? Are there any other changes that I need to make?

Yes, it is considered, definitely! I just haven't had the time to give it a proper review so far.

hishamhm:
> A couple pieces of my code could probably be made cleaner or more readable, but I had to leave them as-is to avoid significantly restructuring code unrelated to the tuple table interface addition (it would feel rude to forcefully inject my personal code tastes into places it isn't needed). I will leave the decision on that up to the maintainers who will actually be interacting with it.

Also, if you'd like to do further code cleanups/changes, feel free to open a PR with them, and we can discuss whether to adopt them there!

PR #1102, opened 2026-03-12
Experimental macro require v2

Frityet:
An update to #1096 with new semantics:

````lua
-- macros.tl
local record macros end
macro macros.double!(x: Statement)
   return ```
      $x
      $x
   ```
end

return macros
````
```lua
-- main.tl
local m = require("macros")
-- also supported: local type m = require("macros")
m.double!(print("hi"))
```


Frityet:
@hishamhm I can also make it into `local macro ... = require(...)`

mtdowling:
I would _love_ exported macros in the next version of Teal. Let me know if there's anything I can do to help.

Frityet:
> I would _love_ exported macros in the next version of Teal. Let me know if there's anything I can do to help.

It all works pretty great right now (I think this PR doesn't have attaching macros to records in it, but that's fixed in my `tl-ng` branch) it's just a matter of agreeing on design. I tried tp make it model off of macroexps, where macroexps can be attached to records and interfaces

Issue #532, opened 2022-06-25
Definition Files for Libraries that Monkey-Patch


tooolbox:
I'm using a Lua module that modifies `string` to add additional functions for handling utf8.  I would like to write some Teal definition files so my linter doesn't complain when I use these functions.

```lua
global record string
    utf8lower: function(string): string
end
```

The above works in that Teal knows my new function is legit, but it blows away the normal definition of `string`, so I am now in a worse situation because all the normal methods are unknown.  What I really need is something like this:

```lua
// utf8-things.d.tl
string.utf8thing = function(string): string

// or
global record extension string
    utf8lower: function(string): string
end
```

I read in [the tutorial](https://github.com/teal-language/tl/blob/master/docs/tutorial.md#the-teal-standard-library-and-lua-compatibility) that the stdlib is declared as const, but I'm not sure the motivation for this or if it's just a matter of no syntax.  Monkey-patching in Lua seems to be a thing, and it would be handy to be able to express this in Teal definition files.  Not necessarily for crazy meta stuff, but for adding/replacing methods.

hishamhm:
> I read in [the tutorial](https://github.com/teal-language/tl/blob/master/docs/tutorial.md#the-teal-standard-library-and-lua-compatibility) that the stdlib is declared as const, but I'm not sure the motivation for this

The generated code makes assumptions that depend on the standard library being constant: It automatically adds polyfills using compat-5.3 and local-izes some accesses in the process.

Monkey-patching implementations should work, but a big thing in statically-typed languages is that the types are, well, static :) - so changing the interfaces of records after the fact would break the assumptions made be the compiler.

Teal already is a bit lenient on what it allows for the sake of familiarity with Lua idioms (e.g defining methods in functions as `function Foo.bar(...)` after the `record Foo` definition).

Messing with the contents of the standard library tables isn't a great idea in general (think what happens if multiple libraries you depend on decide to do it in incompatible ways), so I don't think Teal should provide features to promote these patterns.

For the time being, unfortunately if you really need to use a customized definition of standard library tables, you'll have to redefine the whole thing.

lenscas:
Though I agree that libraries shouldn't have a way to actually edit it. I do think there is benefit in teal being able to consume a .d.tl file and going "this is the std now". However, this should ONLY work for projects (so, be done through the tlconfig.lua file or similar) and not for libraries. As in, loading a library should NOT result in the std definitions being changed.

lenscas:
[...]

PR #1104, opened 2026-03-17
feat: implement no_stdlib option

Aire-One:
Hello, this is a new version for the https://github.com/teal-language/tl/pull/961 I started a while back, but lacked understanding/direction to complete the implementation.

Thank you, @hishamhm for the work on decoupling the stdlib from the default types, and for your latest messages on https://github.com/teal-language/tl/issues/532 that gave me the missing pointers I needed.

This PR adds a new option `no_stdlib` (exposed to the user by CLI arg and by tlconfig field) to disable the stdlib loading by skipping `load_precompiled_default_env`.

Notes: In combination with `global_env_def`, this effectively allow redefining the stdlib. By itself, I think this solves most problems related to using Teal in custom/embedded environments. For next steps, we could allow controlling the precompiled `default_env` by either exposing it to user, or adding new `teal/default/*.d.tl` and `teal/precompiled/` definitions.

### DIFF
--- a/teal/environment.tl
+++ b/teal/environment.tl
@@ -143,6 +143,10 @@ function environment.new(opts?: EnvOptions): Env
    env.opts = opts or env.opts
    if not env.opts.no_stdlib then
       load_precompiled_default_env(env)
+   else
+      -- We still need to load the prelude, which contains language fundamental
+      -- definitions that should always be available.
+      env:require_module("teal.default.prelude")
    end
    return env
 end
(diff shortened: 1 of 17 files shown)


teal/environment.tl
```diff
@@ -140,7 +141,9 @@ end
 function environment.new(opts?: EnvOptions): Env
    local env = empty_environment()
    env.opts = opts or env.opts
-   load_precompiled_default_env(env)
+   if not env.opts.no_stdlib then
+      load_precompiled_default_env(env)
+   end
```
hishamhm: Can you add an `else` section here that loads only the default prelude? Otherwise I think `--no-stdlib` risks becoming a footgun when people fail to require that from their alternative stdlib implementations.

I believe `env:require_module("teal.default.prelude")` should suffice!

PR #1113, opened 2026-04-17
Add missing files to rockspec

euclidianAce:
Attempting to have cyan use `teal.api.v2` showed that some modules were not included in the rockspec.

I also left the shell+awk invocation I used to regenerate the list as a comment for future ref. (my awk skills are rusty, and maybe sed could have done it as well, but it gets the job done)


hishamhm:
@euclidianAce Thank you!!

PR #1122, opened 2026-05-18
Add Lua 5.5 as a target

mbartlett21:
Which means that functions with named varargs will no longer have an extra `local x = table.pack(...)`, and instead use the builtin functionality.

----

I also did some test cleanup, so that the `util.gen` code now actually checks the output, rather than parsing and regenerating (which previously meant that syntax-only compatibility could only be tested with the cli)


hishamhm:
Cool improvements!!

Issue #1032, opened 2025-09-26
Support declaration merging for global records


farism:
Allowing for the extension of global records would be handy, e.g.

```tl
-- some_library.tl
global record my_record
  func_a: function()
end

-- my_app.tl
global record my_record
  func_b: function()
end
```

Right now it doesn't error if a global record is defined twice, but it only picks up the first one it finds.

PR #1127, opened 2026-05-30
Add detection of conflicting global type redefinitions

### LINKED ISSUE #1126: Redeclaring a `global` type alias crashes the compiler
Declaring the same `global` type alias twice crashes `tl` with `attempt to index a nil value (local 'added')`

Example:
```teal
global record Widget
  id: integer
end

global type Handle = Widget
global type Handle = Widget
```

tyulpan:
Initially, I ran into the crash described in #1126. 
My first idea was to simply ignore duplicate declarations. That would be fine for identical types. 

But what if we wanted to do something like this:

```teal
global record Widget
  id: integer
end

global record Gadget
  name: string
end

global type Handle = Widget
global type Handle = Gadget
```

Simply ignoring the duplicate declaration would be a footgun here. 

And that's where I ran into another problem. Related to #1032. Currently, redeclaring a `global` type (`record`, `interface`, or `type` alias) with a different definition isn't reported as an error, and the compiler keeps the first one it finds.

So let's fix that.

Resolves #1126.

### DIFF
--- a/teal/check/relations.tl
+++ b/teal/check/relations.tl
@@ -425,6 +425,11 @@ relations.eqtype_relations = {
       end,
       ["emptytable"] = compare_true_inferring_emptytable_if_not_userdata,
    },
+   ["typedecl"] = {
+      ["typedecl"] = function(ck: TypeChecker, a: TypeDeclType, b: TypeDeclType): boolean, {Error}
+         return ck:same_type(a.def, b.def)
+      end,
+   },
    ["function"] = {
       ["function"] = function(ck:TypeChecker, a: FunctionType, b: FunctionType): boolean, {Error}
          local argdelta = a.is_method and 1 or 0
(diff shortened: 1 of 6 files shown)


teal/check/relations.tl
```diff
@@ -425,6 +425,11 @@ relations.eqtype_relations = {
       end,
       ["emptytable"] = compare_true_inferring_emptytable_if_not_userdata,
    },
+   ["typedecl"] = {
+      ["typedecl"] = function(ck: TypeChecker, a: TypeDeclType, b: TypeDeclType): boolean, {Error}
+         return ck:same_type(a.def, b.def)
```
hishamhm, line 430: The one thing I need to check about this PR is: are we sure that adding this does not accidentally introduce structural type compatibility to an otherwise mostly-nominal type system?

PR #1130, opened 2026-06-07
fix: preserve line comment at end of file without trailing newline

efredriksson:
Found through fuzzing that lexer can drop line comments if nothing comes after it.


hishamhm:
Thanks!

Issue #1121, opened 2026-05-17
Invalid goto causes tl to generate an internal compiler error.


parke:
```
goto  finish
local  _foo  =  0
:: finish ::
local  _bar  =  0
````

When I `tl run` the above script:

```
Internal Compiler Error: Teal generator produced invalid Lua. Please report a bug at https://github.com/teal-language/tl

test.tl:4: <goto finish> at line 1 jumps into the scope of local '_foo'
```

The above code (which is syntactically valid in both Lua and Teal) causes a compile time error in Lua.  The reason is obvious.

However, `tl` asked me to report the error here on GitHub.  So here is my report!

**Comment by hishamhm** [comment]
Thank you for the report! I added the check to the Teal compiler itself, no longer relying on Lua to trigger the error. Should be fixed in `main`!

**Comment by mbartlett21** [comment]
@hishamhm 
That now breaks valid code:

```lua
for i = 1, 5 do
  goto next
  local _this = i
  ::next::
end
```

I did have a different implementation at https://github.com/mbartlett21/tl/tree/mjb/goto-scoping that adds a check for the case of a label not having any non-label statement between it and the end of the block.

EDIT:

Reading through your implementation that seems like it is a bit simpler than mine. It just needs the check that the label is either the last statement in the block, or that it only has labels between it and the end of the block, but I wasn't sure how to do that while visiting the label statement.

mtdowling:
Yeah this broke my project too. I've had to pin to a specific commit.

hishamhm:
Sorry about the mess, @mtdowling's fix is merged now.

PR #1139, opened 2026-07-08
fix: allow goto to a label at the end of a block

mtdowling:
The goto validity check from #1121 rejected jumps over local declarations to a label at the end of a block, such as the common `goto continue` idiom. Lua permits this. A label followed only by other labels at the end of a block is outside the scope of the block's locals.

The relaxation does not apply to labels at the end of a repeat body, whose locals remain visible in the until condition, nor to labels followed by a return statement, matching PUC-Lua.

### DIFF
--- a/spec/lang/statement/goto_spec.lua
+++ b/spec/lang/statement/goto_spec.lua
@@ -141,4 +141,55 @@ describe("goto", function()
    ]], {
       { y = 2, msg = "goto jumps into scope of a local variable" }
    }))
+
+   it("accepts a goto over a local to a label at the end of a block", util.check([[
+      for i = 1, 5 do
+         goto next
+         local _this = i
+         ::next::
+      end
+   ]]))
+
+   it("accepts the continue idiom with locals in the loop body", util.check([[
+      for i = 1, 3 do
+         if i == 2 then
+            goto continue
+         end
+         local doubled = i * 2
+         print(doubled)
+         ::continue::
+      end
+   ]]))
+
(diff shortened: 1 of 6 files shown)


mtdowling:
CI failed due to the playground timing out. Seems to be a recurring issue based on the other open PR.

mtdowling:
I think just merging this would make installing Teal with `--dev` work for me. 🙏

spec/lang/statement/goto_spec.lua
```diff
@@ -141,4 +141,55 @@ describe("goto", function()
...
+         end
+         local doubled = i * 2
+         print(doubled)
+         ::continue::
+      end
+   ]]))
+
+   it("accepts a goto over a local to stacked labels at the end of a block", util.check([[
+      do
+         goto second
```
mbartlett21: This should be `goto first` to actually test what it is saying.

mbartlett21 (approved):
Looks fine to me. Only comment would be on the mentioned test.

Looks like a bit of a cleaner implementation than I had done.

spec/lang/statement/goto_spec.lua
mtdowling: Fixed!

hishamhm (approved):
Thank you!

mtdowling:
Thanks. I think the dev rock needs to be republished? It's still pulling down a version of Teal that keeps says `module 'teal' not found`

hishamhm:
done, thanks for the heads up!

PR #1148, opened 2026-07-18
fix: include source_dir when checking modules

### LINKED ISSUE #1119: `tl check` does not add source_dir to include_dir in tlconfig.lua
**I must note that this is only a bug on `tl check`. `cyan build` works perfectly fine with this setup**.

```teal
-- src/vec.tl
local record Vec
  x: number
  y: number
end

return function(x, y): Vec
  return {x=x, y=y} as Vec
end
```

```teal
-- src/field.tl
local vec = require "vec"
--                  ^ module not found: 'vec'
```

```lua
-- tlconfig.lua
return {
  include_dir = {"include"}, -- Can't remove, I'm using an external library
  source_dir = "src",
  build_dir = "build"
}
```

If I replace the `require "vec"` with a `require "src.vec"`, the error goes away in the editor, but the build starts to fail.

The temporary fix I attempted (adding "src" to include_dir manually) resulted in a warning.

floze-the-genius:
## Summary

- recognize `source_dir` as a validated string option in `tlconfig.lua`
- add `source_dir` to the effective module search path while preserving configured `include_dir` entries and CLI `-I` precedence
- add a CLI regression covering imports from both `source_dir` and `include_dir`

Fixes #1119.

## Testing

- `busted --suppress-pending -v spec/cli/include_dir_spec.lua` (3 successes)
- `luarocks lint tl-dev-2.rockspec`
- `make selfbuild`
- `luarocks test` (1924 successes, 0 failures, 9 pending)
- manual reproduction from #1119 with `source_dir = "src"` and a separate `include_dir`

## AI assistance

This contribution was implemented with assistance from OpenAI Codex. The reported failure was reproduced before the change, and the focused regression plus the full upstream test suite were run after it.

### DIFF
--- a/docs/src/compiler_options.md
+++ b/docs/src/compiler_options.md
@@ -22,6 +22,7 @@ return {
 | -------------------- | -------------------------- | ---------- | -------------------- | ----------- |
 | `-l --require`       |                            | `{string}` | `run`                | Require a module prior to executing the script. This is similar in behavior to the `-l` flag in the Lua interpreter. |
 | `-I --include-dir`   | `include_dir`              | `{string}` | `check` `gen` `run`  | Prepend this directory to the module search path.
+|                      | `source_dir`               | `string`   | `check` `gen` `run`  | Set the project source directory and add it to the module search path.
 | `--gen-compat`       | `gen_compat`               | `string`   | `gen` `run`          | Generate compatibility code for targeting different Lua VM versions. See [below](#generated-code) for details.
 | `--gen-target`       | `gen_target`               | `string`   | `gen` `run`          | Minimum targeted Lua version for generated code. Options are `5.1`, `5.3` and `5.4`. See [below](#generated-code) for details.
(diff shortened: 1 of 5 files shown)


hishamhm:
@euclidianAce Could you give this small PR a quick review? Just want to make sure that any new behavior added is compatible with Cyan's.

euclidianAce:
I'd like a small blurb in [docs/compiler_options.md](https://github.com/teal-language/tl/blob/a2997d589a291427e9d9dd51b45fd9eb307a7e5e/docs/src/compiler_options.md) explaining what the option is and why it exists. Other than that looks fine to me.

floze-the-genius:
Added the requested `source_dir` documentation in `030e7c2`: the compiler-options table now lists it, and the text explains why it participates in module resolution while preserving explicit `include_dir` entries.

Validation on the updated head:
- `mdbook build` succeeded
- repository cspell command checked 39 Markdown files with 0 issues
- `git diff --check` passed

The previous Node 18 build failure was a Netlify authorization timeout rather than a project build or test failure; the new head should retrigger the PR checks.

PR #1150, opened 2026-07-23
Fix incorrect `math.mininteger` polyfill

purplesyringa:
`-2^53 - 1` is not a valid integer in Lua 5.2, which didn't have true integer support and thus relied on floats. You can verify that it's not an integer because `-2^53 - 1` compares equal to `-2^53`. `-2^53` is the smallest integer that is actually distinguishable from every other integer up until `math.maxinteger`.

`math.maxinteger` is fine.


hishamhm:
@purplesyringa Great catch! Thank you!

PR #1155, opened 2026-08-09
feat: add luacats generation

ALameLlama:
Creating this as a draft PR so people can look at it, I've been playing around with different versions of this for a while.

This creates the command gen arg/config `emit-luacats` which will create luacat comments for functions params/returns, union, records enums etc.

This is mostly working and I'll try and create specs for this so there will be tests to ensure it's working, namespaces are still gonna be a pain especially if it's cross package so that is something I'll need to think about too.

As a POC, I cloned https://github.com/FourierTransformer/tinytoml/compare/main...ALameLlama:tinytoml:luacats-gen and ran `tl gen tinytoml.tl --emit-luacats`

I'm also not sure if this should be `emit-luacats` or `gen-luacats` since internally we use `gen` a little differently

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ SOURCES = teal/debug.tl teal/attributes.tl teal/errors.tl teal/lexer.tl \
 	teal/check/type_checker.tl teal/check/node_checker.tl \
 	teal/input.tl \
 	teal/check/require_file.tl \
-	teal/gen/targets.tl teal/gen/lua_generator.tl teal/gen/lua_compat.tl \
+	teal/gen/targets.tl teal/gen/luacats.tl teal/gen/lua_generator.tl teal/gen/lua_compat.tl \
 	teal/package_loader.tl teal/loader.tl \
 	teal/api/v2.tl teal/api/v1.tl \
 	teal/init.tl \
--- a/docs/src/compiler_options.md
+++ b/docs/src/compiler_options.md
@@ -25,6 +25,7 @@ return {
 |                      | `source_dir`               | `string`   | `check` `gen` `run`  | Set the project source directory and add it to the module search path.
 | `--gen-compat`       | `gen_compat`               | `string`   | `gen` `run`          | Generate compatibility code for targeting different Lua VM versions. See [below](#generated-code) for details.
 | `--gen-target`       | `gen_target`               | `string`   | `gen` `run`          | Minimum targeted Lua version for generated code. Options are `5.1`, `5.3` and `5.4`. See [below](#generated-code) for details.
+| `--emit-luacats`     | `emit_luacats`             | `boolean`  | `gen`                | Generate LuaCATS annotations from explicit Teal type declarations.
 | `--keep-hashbang`    |                            |            | `gen`                | Preserve hashbang line (`#!`) at the top of file if present.
 | `-p --pretend`       |                            |            | `gen`                | Don't compile/write to any files, but type check and log what files would be written to.
 | `--wdisable`         | `disable_warnings`         | `{string}` | `check` `run`        | Disable the given warnings.
--- a/teal.lua
+++ b/teal.lua
@@ -11356,10 +11356,12 @@ local types = require("teal.types")
 
 
 local traversal = require("teal.traversal")
+local luacats = require("teal.gen.luacats")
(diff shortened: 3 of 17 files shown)


lenscas:
What does this output for generic types? Considering that Luna language server doesn't support generic types.

ALameLlama:
Something like
```
local function generic_type_conversion<T>(raw_value: string): T return raw_value as T end
```

Will generate something like this
```
---@generic T
---@param raw_value string
---@return T
local function generic_type_conversion(raw_value) return raw_value end
```

Over the next week or two I'll try and get a bunch of tests written up for all the different types that luacats supports https://luals.github.io/wiki/annotations/

lenscas:
I meant a generic record/interface.
```
local record Foo<T>
    a: T
end
```

what would be generated for that? As emitting a `---@generic T`  will cause problems for lua language server, which as far as I'm aware is the only consumer of this annotation. Thus emitting something that doesn't work for it sounds... counter productive...

ALameLlama:
Oh sorry, It would generate this.

```teal
local record Bar<T>
    a: T
end

local bar1: Bar<string> = {
    a = "hello"
}

local bar2: Bar<number> = {
    a = 42
}

print(bar1.a)
print(bar2.a)
```

would generate
```lua
---@class Bar<T>
---@field a T

---@type Bar<string>
local bar1 = {
	a = "hello",
}

---@type Bar<number>
local bar2 = {
	a = 42,
}

print(bar1.a)
print(bar2.a)
```

Yeah that's fair point on the lua ls side, they do have an open issue for tracking generics.

I like to use this when working with lua https://github.com/emmyluals/emmylua-analyzer-rust which does support [generics](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/main/docs/emmylua_doc/annotations_EN/README.md#usage-guide)

e.g
```teal
local record Foo<T>
    a: T
end

local function make_foo<T>(value: T): Foo<T>
    return {
        a = value
    }
end

local function get_value<T>(foo: Foo<T>): T
    return foo.a
end

local string_foo: Foo<string> = make_foo("hello")

local good: string = get_value(string_foo)
```
would generate
```lua
---@class Foo<T>
---@field a T

---@generic T
---@param value T
---@return Foo<T>
local function make_foo(value)
	return {
		a = value,
	}
end

---@generic T
---@param foo Foo<T>
---@return T
local function get_value(foo)
	return foo.a
end

---@type Foo<string>
local string_foo = make_foo("hello")

---@type string
local good = get_value(string_foo)
```

and in the lua side if I added and ran emmylua_check
```lua
---@type number
local bad = get_value(string_foo)
```

```bash
> emmylua_check generic.lua
---  [1 warning]
warning: Cannot assign `string` to `number`.  [assign-type-mismatch]
  -->: :27:7

  26 │ ---@type number
  27 │ local bad = get_value(string_foo)
  28 │



Summary
  1 warning

Check completed with warnings
Check finished
```


Ideally I'd like to be able to write packages/code in teal, generate the lua code and have people that don't use teal still have a good experience with the package since it would have annotation.

hishamhm:
> Ideally I'd like to be able to write packages/code in teal, generate the lua code and have people that don't use teal still have a good experience with the package since it would have annotation.

That is a nice sentiment, but differences in type systems go beyond syntax. For example, Teal has its own interface model (with `where` and `is` semantics) that I don't expect to be directly mappable to other implementations of typed Lua extensions. Transpiling the syntax of one type system into another without ensuring that the semantics match could become a minefield for end-users, once you step beyond the basics. 

If someone wants to make a third-party tool that does this conversion, at their own risk, it's their choice to do so, but I don't think we should add exporting of other type system formats into the core Teal compiler, otherwise every difference between them, present and future (i.e. every time either we or LuaCATS makes a change), is going to become a new bug report sent our way, and the burden of maintaining and fixing them will ultimately fall on our side.

We already cover the costs of supporting multiple `gen-targets` for the various Lua dialects. We don't have the resources to also support multiple typed Lua dialects, sorry!

ALameLlama:
> We already cover the costs of supporting multiple gen-targets for the various Lua dialects. We don't have the resources to also support multiple typed Lua dialects, sorry!

Honestly that's completely fair

PR #1156, opened 2026-08-10
fix: apply Lua compatibility code in tl.loader() and tl.load()

efredriksson:
I ran into this while I set up a job to test Cerulean with Teal dev.

I am not 100% sure if this is the right fix. Other alternatives:
1. Call `lua_compat.apply` inside `input.check` or at the end of `check.check`, which
   restores the code to be closer to how it was in 0.24.8.
2. Thread `gen_compat` through `lua_generator.generate`, which changes its signature.

If you want I could change to either or open an issue instead.

### DIFF
--- a/spec/api/v2/load_spec.lua
+++ b/spec/api/v2/load_spec.lua
@@ -1,5 +1,7 @@
 require("compat53")
 
+local util = require("spec.util")
+
 describe("tl.load", function()
 
    describe("loading Teal code from Lua", function()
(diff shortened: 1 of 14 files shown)


hishamhm:
This approach seems fine!

I did notice however that the other places where `lua_compat.apply` is called is always wrapped by stuff like `if result and result.ast then`. We should probably either add that to these new calls or, perhaps even better, move that check inside `apply` and simplify all call sites.

hishamhm:
(Regarding the other approaches you suggested, I chose to not have the `check` calls modify the shape of the tree anymore.)

efredriksson:
I added a new commit with the refactor, please have a look.

hishamhm:
Looks great, thank you!!

PR #1157, opened 2026-08-14
Fix compilation of Teal scripts under Lua 5.5

mszaro:
Teal does not run on Lua 5.5. 

Lua 5.5 makes the first variable of a generic `for` loop `const`, which broke tlcli/driver.tl. So the compiler cannot load itself. That is a one line fix. I regenerated the committed .lua with make selfbuild.

The second is tl run. It compiled for the 5.3 target, then loaded the result into the running VM. The shadow that allows assigning to a for-in variable is only emitted for 5.4 and 5.5. So any script using that feature died on 5.5 with an internal compiler error. It now takes its target from teal.runtime_target(), like tl.load already does.

Tests pass on 5.5.0 and 5.4.8. Macro bodies still generate for 5.1 and skip the compat pass, so that path is still broken.

### DIFF
--- a/spec/cli/run_spec.lua
+++ b/spec/cli/run_spec.lua
@@ -28,6 +28,23 @@ describe("tl run", function()
          ]], output)
       end)
 
+      it("runs code that modifies a for loop control variable", function()
+         local name = util.write_tmp_file(finally, [[
+            local t: {string:string} = { a = "1" }
+
+            for k, v in pairs(t) do
+               k = k:upper()
+               print(k .. "=" .. v)
+            end
+         ]])
+         local pd = io.popen(util.tl_cmd("run", name), "r")
+         local output = pd:read("*a")
+         util.assert_popen_close(0, pd:close())
+         util.assert_line_by_line([[
+            A=1
+         ]], output)
+      end)
+
       it("reports number of errors in stderr and code 1 on type errors", function()
          local name = util.write_tmp_file(finally, [[
             local function add(a: number, b: number): number
--- a/docs/src/compiler_options.md
+++ b/docs/src/compiler_options.md
@@ -90,6 +90,9 @@ generating 5.3-compatible code. If you install `tl` using LuaRocks, the CLI
 will use the Lua version you use with LuaRocks, so it will default to that
 Lua's version.
 
+The `tl run` command loads the code it generates into the Lua VM it is running
+on, so unless you set `--gen-target` it targets that exact version.
+
 If you require the `tl` Lua module and use the `tl.loader()`, it will do the
 implicit version selection, picking the right choice based on the Lua version
 you're running it on.
--- a/tlcli/commands/run.lua
+++ b/tlcli/commands/run.lua
@@ -37,6 +37,15 @@ local function type_check_and_load(tlconfig, filename)
 end
 
 return function(tlconfig, args)
+
+   local runtime_target = teal.runtime_target()
+   if runtime_target and not tlconfig["gen_target"] then
+      tlconfig["gen_target"] = runtime_target
+      if runtime_target == "5.4" or runtime_target == "5.5" then
+         tlconfig["gen_compat"] = "off"
+      end
+   end
+
    if args["require"] then
       tlconfig._init_env_modules = {}
       for _, module in ipairs(args["require"]) do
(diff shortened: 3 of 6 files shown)


spec/cli/run_spec.lua
```diff
@@ -28,6 +28,23 @@ describe("tl run", function()
          ]], output)
       end)
 
+      it("runs code that modifies a for loop control variable", function()
+         local name = util.write_tmp_file(finally, [[
+            local t: {string:string} = { a = "1" }
+
+            for k, v in pairs(t) do
+               k = k:upper()
```
Aire-One, line 36: I don't get how this test can succeed with Lua 5.5 ?

mszaro: That's true, it's illegal in Lua 5.5. But this is some Teal code as a string literal.

So that is what this test is doing, it is Teal code that reassigns `k` inside the loop body (legal in Teal but not in Lua 5.5), which is then compiled with `tl run`, and then verified that the Lua runtime actually accepts and executes it correctly. 

Ergo Teal's output declares a shadow local at the top of the loop body `local k = k`, so the assignment targets that rather than the const control variable.

Aire-One: > Ergo Teal's output declares a shadow local at the top of the loop body local k = k, so the assignment targets that rather than the const control variable.

Ok! That's the part I was missing. After testing, I can confirm `./tl gen --gen-target 5.5 --gen-compat off test-for-loop-55.tl` did indeed generate the local copy for k. 

Sorry about that.

(With that in mind, is the change in `tlcli/driver.tl` still needed, tho?)

mszaro: I believe it is, but I am a first time contributor so I can't claim to be intimately familiar with the code base.

Here's where my logic takes me: That local k = k trick only happens for the 5.4 and 5.5 targets. driver.tl doesn’t go through tl run at all, it gets built by make selfbuild, and the Makefile pins that to --gen-target=5.1 (TLGENFLAGS = --check --gen-target=5.1). So, I wasn't 100% sure that this change would be reliably exercised in the actual teal compiler itself.

Aire-One: Understood. Thank you!

hishamhm:
Hi @mszaro, thanks for your PR!

> Here's where my logic takes me: That local k = k trick only happens for the 5.4 and 5.5 targets. driver.tl doesn’t go through tl run at all, it gets built by make selfbuild, and the Makefile pins that to --gen-target=5.1 (TLGENFLAGS = --check --gen-target=5.1). So, I wasn't 100% sure that this change would be reliably exercised in the actual teal compiler itself.

That is correct, and it brings us to a tricky point. I'd really rather not code `tlcli` worrying about the underlying Lua version — that is, after all, one of the things that Teal aims to adress. I've spent many years coding LuaRocks using the "lowest common denominator" of all Lua versions, and that's no fun.

But because of `tlcli`'s "bootstrap" position in the build chain, I understand that just editing the Makefile to bump it to `--gen-target=5.5` wouldn't cut it. I'm thinking of some alternatives:

* I'm wondering now if we should enable the `local k = k` codegen trick to all gen-targets; that would be probably the easiest thing in order to improve compatibility, as a lot more code built with `--gen-target=5.1` would become immediately Lua 5.5 compatible. But then that would encourage people to `--gen-target=5.1` and then hope for the best in terms of compatibility, which [holds people back to old versions](https://hisham.hm/2026/08/04/the-lua-community-needs-to-learn-to-move-on/) while still running sub-optimal code (e.g. compat code instead of `//`, etc.).
* Or maybe we should just build the Lua sources for all supported versions, and have users use the right one. That's not something that LuaRocks currently supports easily, but it's not impossible: we could publish one rockspec for each supported Lua version (e.g. `tl 0.25.0.51`, `tl 0.25.0.52`, etc. and based on the `lua` dependency, each would appear in its own LuaRocks manifest — we've done such things for other Lua packages before).
* Or maybe it's time to really add first-class Teal support in LuaRocks, so that people can ship only `.tl` files and then LuaRocks can compile them to the right gen-target based on the user's Lua version at installation time.

While I don't reach a decision, I'll gladly take your PR with the change to `tlcli/driver.tl` as an interim solution!

mszaro:
> I'm wondering now if we should enable the local k = k codegen trick to all gen-targets

To me, that sounds like a reasonable alternative if you want to reduce conditionality and complexity. There's no harm in doing so for older versions - I just didn't want to assume. Thanks @hishamhm!

PR #1159, opened 2026-08-18
Some documentation updates for Lua 5.5

bjornbm:
Mainly update to parts of the documentation to reflect that Lua 5.5 is a compiler target. Rather low hanging fruit I will admit.

Someone should confirm that “The options `5.4` and `5.5` are equivalent to `5.3`, but also allow using the `<close>`
variable annotation.” (i.e., `5.5` does not introduce other changes that deserve mention. I admit I did not dig deep.)

### DIFF
--- a/docs/src/functions.md
+++ b/docs/src/functions.md
@@ -159,8 +159,6 @@ local function test(...args: boolean): integer, boolean
 end
 ```
 
-Note, however, that named vararg tables will not be read-only when targeting Lua <5.5.
-
 In case your function returns a variable amount of values, you may also declare
 variadic return types by using the `type...` syntax:
 
--- a/docs/src/functions.md
+++ b/docs/src/functions.md
@@ -150,6 +150,17 @@ end
 test(1, 2, 3)
 ```
 
+Named vararg tables, as introduced in [Lua 5.5](https://www.lua.org/manual/5.5/manual.html#3.4.11),
+are supported by Teal:
+
+```lua
+local function test(...args: boolean): integer, boolean
+   return args.n, args[1]
+end
+```
+
+Note, however, that named vararg tables will not be read-only when targeting Lua <5.5.
+
 In case your function returns a variable amount of values, you may also declare
 variadic return types by using the `type...` syntax:
 
--- a/docs/src/functions.md
+++ b/docs/src/functions.md
@@ -38,7 +38,7 @@ argument type itself.
 
 You can declare functions that generate iterators which can be used in
 `for` statements: the function needs to produce another function that iterates.
-This is an example [taken the book "Programming in Lua"](https://www.lua.org/pil/7.1.html):
+This is an example [taken from the book "Programming in Lua"](https://www.lua.org/pil/7.1.html):
 
 ```lua
 local function allwords(): (function(): string)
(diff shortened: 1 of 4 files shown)


hishamhm:
Thanks! This is much appreciated!

Perhaps a "non-difference" that might be worth mentioning in the documentation is that Teal does allow modifiying the `for` index variables (like Lua 5.4 and below) even when the target is Lua 5.5.

hishamhm (approved):

bjornbm:
> Perhaps a "non-difference" that might be worth mentioning in the documentation is that Teal does allow modifiying the `for` index variables (like Lua 5.4 and below) even when the target is Lua 5.5.

Where would be a suitable place for this? https://github.com/teal-language/tl/blob/main/docs/src/teal_standard_library_and_lua_compatibility.md?

Also, if this is the desired behavior in Teal code, I think that when the target is 5.5 (or any target that should be compatible with 5.5) the following transformation should be used to avoid a crash (`attempt to assign to const variable 'i'`) on 5.5:

```lua
for i = 1, 5 do
   i = i + 1
   print(i)
end
```

Should become:

```lua
for i = 1, 5 do local i = i
   i = i + 1
   print(i)
end
```

(Deserves a new issue if you agree)

hishamhm:
@bjornbm  the compiler already does this :)

bjornbm:
@hishamhm I believe I am seeing the numeric for exception reported as a one of the incidental observations in #1158:

> ## Not addressed here
> 
> Two things I ran into while testing, both out of scope:
> 
> **Numeric `for` has the same problem, and it is worse.** Lua 5.5 also makes the
> control variable of a numeric `for` a constant, and `lua_compat.adjust_code`
> has no `fornum` case at all. So ordinary, non-macro Teal code that writes to a
> numeric `for` variable type checks cleanly and then produces invalid Lua on 5.5:
> 
> ```
> $ lua5.5 ./tl run fornum.tl
> Internal Compiler Error: Teal generator produced invalid Lua. Please report a
> bug at https://github.com/teal-language/tl
> 
> fornum.tl:3: attempt to assign to const variable 'i'
> ```
> 
> This reproduces on master without any of the changes here and affects code that
> has nothing to do with macros. It needs `lua_compat` and `visitors` changes, so
> I left it alone — happy to open a separate issue for it.

I tried with an “pairs”-for and then the a local copy of the key is indeed assigned if the target is 5.4 or 5.5. So it seems the numeric for case was forgotten.

docs/src/functions.md
```diff
@@ -150,6 +150,17 @@ end
...
+Named vararg tables, as introduced in [Lua 5.5](https://www.lua.org/manual/5.5/manual.html#3.4.11),
+are supported by Teal:
+
+```lua
+local function test(...args: boolean): integer, boolean
+   return args.n, args[1]
+end
+```
+
+Note, however, that named vararg tables will not be read-only when targeting Lua <5.5.
```
hishamhm: Let's not document this line here. Instead, let's open an issue to get this fixed.

bjornbm: Note removed. We have #1160 for this since yesterday.

hishamhm: Ahh sure, sorry!

PR #1165, opened 2026-08-26
compat: shadow numeric for control variables on 5.4+

### LINKED ISSUE #1164: --gen-target=5.5 emits invalid Lua for numeric for loops that assign to the control variable
Follow-up to the discussion in #1159. #1158 covered the macro-body path; this
is the ordinary, non-macro one.

Lua 5.5 makes the control variable of a numeric `for` a constant, the same as
the generic `for`:

    $ lua5.5 -e 'for i=1,2 do i = 1 end'
    lua5.5: attempt to assign to const variable 'i'

`lua_compat.adjust_code` inserts the shadowing local for `forin`, but there is
no `fornum` case, so this type checks and then fails at runtime:

```lua
-- fornum.tl
local t = { 10, 20 }
for i = 1, #t do
   i = i + 1
   print(i)
end
```

    $ tl check fornum.tl
    0 errors detected

    $ lua5.5 ./tl run fornum.tl
    Internal Compiler Error: Teal generator produced invalid Lua.
      fornum.tl:3: attempt to assign to const variable 'i'

There are two halves to it: `lua_compat` has no `fornum` visitor, and there is
no numeric counterpart to `forin_modifies_control_var`, which the type checker
sets in `check/visitors.tl`.

Since you mentioned documenting that Teal permits modifying `for` index
variables even when targeting 5.5, this looks like the numeric case not
delivering on that.

Two ways it could go, and I wasn't sure which you'd prefer:

1. Mirror the `forin` shadow — add the visitor and the flag.
2. Make the numeric control variable const in the type checker for all
   targets, like the named-vararg decision. Assigning to a loop counter is
   arguably always a mistake.

Happy to send a PR once the direction is clear.

JJH090501:
Closes #1164. Option 1, as decided in the issue: mirror the `forin` shadow
rather than making numeric control variables const in all targets.

## Problem

`lua_compat.adjust_code` only had a `forin` case, so plain (non-macro) code that
writes to a numeric `for` variable type checks and then generates invalid Lua:

```lua
local t: {string} = { "a", "b", "c" }

for i = 1, #t do
   i = i + 0
   print(t[i])
end
```

```
$ lua5.5 ./tl run fornum.tl
Internal Compiler Error: Teal generator produced invalid Lua. Please report a bug at https://github.com/teal-language/tl

fornum.tl:3: attempt to assign to const variable 'i'
```

## Change

- `teal/ast.tl` — `fornum_modifies_control_var` on `Node`.
- `teal/check/visitors.tl` — the `fornum` visitor sets it from
  `has_been_written_to`, the same way `forin` does, so writes from nested
  functions are covered too.
- `teal/gen/lua_compat.tl` — a `fornum` case. The shadow construction that was
  inlined in the `forin` case is factored out into `shadow_control_var()` and
  shared; the new case is six lines.

The shadow is a no-op in every target, since writing to a numeric `for` variable
has never affected the iteration:

```
$ lua5.1 -e 'for i=1,3 do i = i + 100; io.write(i," ") end'
101 102 103
```

## Tests

`spec/lang/code_gen/fornum_spec.lua`, deliberately parallel to `forin_spec.lua`:
no shadow on 5.3, a shadow on 5.4 when the body assigns, none when it does not,
a loop with an explicit step, an assignment made from a nested function, and
nested loops where only the inner one is written to.

`make selfbuild` clean, `busted --suppress-pending spec/` 1932 → 1938. The
reproduction above was verified end to end on Lua 5.5.1.

### DIFF
--- a/spec/lang/code_gen/fornum_spec.lua
+++ b/spec/lang/code_gen/fornum_spec.lua
@@ -1,0 +1,97 @@
+local util = require("spec.util")
+
+describe("fornum", function()
+   it("5.3: doesn't generate control variable that is local to the iteration", util.gen([[
+      local t: {string} = { "a", "b", "c" }
+
+      for i = 1, #t do
+         i = i + 1
+         print(t[i])
+      end
+   ]], [[
+      local t = { "a", "b", "c" }
+
+      for i = 1, #t do
+         i = i + 1
+         print(t[i])
+      end
+   ]], "5.3"))
+
+   it("5.4: generates control variable that is local to the iteration", util.gen([[
+      local t: {string} = { "a", "b", "c" }
+
+      for i = 1, #t do
+         i = i + 1
+         print(t[i])
+      end
+   ]], [[
+      local t = { "a", "b", "c" }
+
+      for i = 1, #t do local i = i
+         i = i + 1
+         print(t[i])
+      end
+   ]], "5.4"))
+
+   it("5.4: does not generate control variable if not assigned to", util.gen([[
+      local t: {string} = { "a", "b", "c" }
+
+      for i = 1, #t do
+         local j = i + 1
+         print(t[j])
+      end
+   ]], [[
+      local t = { "a", "b", "c" }
+
+      for i = 1, #t do
+         local j = i + 1
+         print(t[j])
+      end
+   ]], "5.4"))
+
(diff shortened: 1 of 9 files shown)


hishamhm:
@JJH090501 Thank you! Rebased and merged!

PR #1166, opened 2026-08-27
ci: add Lua 5.5 to the test matrix

JJH090501:
Lua 5.5 has been out since December, and it is the one version CI does not
build. All four 5.5 issues so far surfaced from user reports rather than from a
failing build.

Teal already passes on it. On Lua 5.5.1 with LuaRocks 3.13.0:

```
make selfbuild                        clean, no diff between stages
busted --suppress-pending spec/       1939 successes / 0 failures / 0 errors
```

Every rock dependency installs, including the C ones (`luafilesystem`,
`luasystem`), and `luarocks.org` serves `manifest-5.5`.

So this is not a fix — it just makes CI check what already works, so the next
regression is caught before a release instead of after one.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
   Test:
     strategy:
       matrix:
-        lua-version: ["5.4", "5.3", "5.2", "5.1", "luajit"]
+        lua-version: [ "5.5", "5.4", "5.3", "5.2", "5.1", "luajit"]
         os: ["ubuntu-latest"]
         include:
         - os: "macos-latest"
--- a/luarocks.lock
+++ b/luarocks.lock
@@ -1,8 +1,8 @@
 return {
    dependencies = {
-      argparse = "0.7.1-1",
-      compat53 = "0.14.4-1",
-      luafilesystem = "1.8.0-1"
+      argparse = "0.7.2-1",
+      compat53 = "0.15.1-1",
+      luafilesystem = "1.9.0-1"
    },
    test_dependencies = {
       luasystem = "0.2.1"


JJH090501:
CI is red, and it found exactly the kind of thing this PR is meant to find —
though not in Teal itself.

```
Forcing argparse to pinned version 0.7.1-1
Error: Could not satisfy dependency argparse 0.7.1-1: No results matching query were found for Lua 5.5.
```

`luarocks.lock` pins `argparse 0.7.1-1`, whose rockspec says
`"lua >= 5.1, < 5.5"`, so it is not in luarocks.org's `manifest-5.5` at all.
argparse fixed this upstream (`rockspec: bump max lua version`, Jan 2026) and
released it as `0.7.2-1` in April. The other pins are likely in the same
position — `compat53 0.14.4-1`, `luafilesystem 1.8.0-1`, and the test pin
`luasystem 0.2.1` — argparse just fails first.

Without the lockfile, everything resolves and passes on 5.5: `make selfbuild`
is clean and `busted --suppress-pending spec/` is 1939/0 on Lua 5.5.1, with the
resolved versions being argparse 0.7.2-1, compat53 0.15.1-1,
luafilesystem 1.9.0-1 and luasystem 0.7.1-1.

So the matrix line is not enough on its own. How would you like to handle it?

1. Bump the pins in `luarocks.lock` to the versions above. Smallest change, but
   it moves the pinned versions for every Lua version, not just 5.5.
2. Skip the lockfile on 5.5 only, until the ecosystem catches up.
3. Something else you have in mind — it's your lockfile, and I don't want to
   guess at the policy.

Happy to push whichever.

(The `playground / build` failure is unrelated — `netlify login` times out
because secrets aren't passed to pull requests from forks.)

hishamhm:
1

JJH090501:
@hishamhm Done — bumped the three pins that block 5.5:

| pin | why |
|---|---|
| `argparse 0.7.1-1` → `0.7.2-1` | rockspec said `lua >= 5.1, < 5.5`, so it is not in `manifest-5.5` |
| `compat53 0.14.4-1` → `0.15.1-1` | same upper bound |
| `luafilesystem 1.8.0-1` → `1.9.0-1` | see below |

`luafilesystem` is the interesting one. Its rockspec has no upper bound, so it
*is* in `manifest-5.5` and resolves fine — but 1.8.0 does not build:

```
src/lfs.c:335:2: error: #error unsupported Lua version
Error: Build error: Failed compiling object src/lfs.o
```

The version ladder in `lfs.c` only goes up to 5.4. 1.9.0 adds 5.5.

I left `luasystem 0.2.1` alone — it compiles and runs on 5.5 as is, so there
seemed no reason to move it.

All three new versions declare `lua >= 5.1` (with `< 5.6` where they have an
upper bound), so they stay valid for every other entry in the matrix. CI will
confirm.

CI is green now.

hishamhm:
Thank you!

PR #1169, opened 2026-09-05
fix: resolve type aliases when lowering 'is' to Lua (#1149)

### LINKED ISSUE #1149: `is` handles type aliases of `integer` incorrectly and `nil` inconsistently
```lua
local x = nil

print(x is integer) -- lowers to math.type(x) == "integer"
print(x is nil) -- lowers to x == nil

local type A = integer
print(x is A) -- lowers to type(x) == "number"
local type B = nil
print(x is B) -- lowers to type(x) == "nil"
```

chiliec:
## What

Closes #1149.

`x is SomeType` compiles to a runtime type check. When the tested type is a **type alias**, the code generator inspected the nominal type's own `typename` (`"nominal"`) instead of the type it resolves to, so `integer` and `nil` aliases were lowered like a generic `type(x) == "..."` check.

```lua
local x: any = nil
local type A = integer
print(x is A)   -- was: type(x) == "number"   (wrong)
local type B = nil
print(x is B)   -- was: type(x) == "nil"      (wrong)
```

`x is A` should lower to `math.type(x) == "integer"` (a `number` may be a float, not an integer), and `x is B` to `x == nil` — exactly what `x is integer` / `x is nil` already produce.

## Fix

In both `is`-lowering sites (`teal/gen/lua_generator.tl` and the 5.1/5.3 compat header selection in `teal/gen/lua_compat.tl`), resolve the cast type through its nominal chain before discriminating on `integer` / `nil`. Chained aliases (`local type C = A`) resolve too.

## Tests

Added a `util.gen` regression test in `spec/lang/operator/is_spec.lua` asserting alias and direct forms lower identically.

Verified genuine RED→GREEN: reverting the source fix (keeping the test) makes it fail with `type(x) == "number"` vs the expected `math.type(x) == "integer"`; restoring it passes.

## Validation

```
make selfbuild            # two-stage bootstrap diff clean
busted spec/lang          # 1733 successes / 0 failures
busted spec/api           # 96 successes / 0 failures
busted spec/cli           # 112 successes / 0 failures
```

Regenerated `.lua` artifacts via `make selfbuild` so the committed build matches the `.tl` sources. Happy to adjust.

### DIFF
--- a/spec/lang/operator/is_spec.lua
+++ b/spec/lang/operator/is_spec.lua
@@ -452,6 +452,23 @@ if not (r == nil) then
    r.f()
 end]]))
 
+      it("lowers 'is' on a type alias like its underlying type (regression test for #1149)", util.gen([[
+         local x: any = nil
+         local type MyInt = integer
+         local type MyNil = nil
+         print(x is MyInt)
+         print(x is MyNil)
+         print(x is integer)
+         print(x is nil)
+      ]], [[
+local x = nil
+
+
+print(math.type(x) == "integer")
+print(x == nil)
+print(math.type(x) == "integer")
+print(x == nil)]], "5.4"))
+
    end)
 
    describe("on while", function()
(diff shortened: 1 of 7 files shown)


hishamhm:
Thank you!

PR #1170, opened 2026-09-15
fix: check the __eq metamethod for '~=', not just '=='

youdie006:
`tl check` reports `argument 2: got Rec, expected number` for `a == b` and reports nothing for `a ~= b`, on the same record with a wrong `metamethod __eq` signature.

The branch at `teal/check/visitors.tl:2124` is entered for `==` **or** `~=`, but it looks the name up as `binop_to_metamethod[node.op.op]`, and that table has `["=="] = "__eq"` and no `~=` entry. The lookup yields nil, and `Context:check_metamethod` indexes with `ameta[method_name or ""]`, so the check is skipped silently rather than raising.

Lua defines `a ~= b` as `not (a == b)` and dispatches the same `__eq`, so a wrong `__eq` signature was only reported if the code happened to spell the comparison `==` somewhere. The two sibling call sites at `:2162` and `:2225` guard with `if mt_name then`; this one passes the nil straight through.

Same shape as #1168, where `~` meant two metamethods and the dispatch went on the token alone.

I used a literal at the call site rather than adding `["~="] = "__eq"` to the shared table, to keep `binop_to_metamethod` unchanged for `gen/lua_compat.tl`. Happy to switch to the table entry if you prefer it.

<details>
<summary>Verification</summary>

Lua 5.4, busted 2.3.0. Generated files come from `make selfbuild`, never hand-edited.

| row | `teal/check/visitors.tl` md5 | `eq_spec.lua` |
|---|---|---|
| main | `439a5fc2ada7cf74e33a768266041a87` | **1 success, 1 failure** at line 18: `Expected 1 type errors, got none` |
| this change | `64a08db11b16fe879ddbc052c05bf856` | 2 successes |
| plain revert to the table lookup | `4acce1bc33e92bfae73095a3fd03991c` | kills the **`~=`** test only |
| `node.op.op == "~=" and "__eq" or nil` | `6bd1a56827030a7e69e92af954f729ab` | kills the **`==`** test only |

The last two kill different tests, so neither spec case is redundant: the revert proves the `~=` row and the over-correction proves the `==` row is load-bearing.

Full suite in one container: **1942 successes / 0 failures / 0 errors** on main, **1944** with this change. The delta is exactly the two added cases.

`spec/lang/metamethods/` had specs for `add`, `bnot`, `call`, `idiv`, `index`, `le`, `len`, `lt`, `tostring` and `unm`, and none for `eq`.

</details>

Disclosure: written with AI assistance (Claude). Every number above is measured.


hishamhm:
@youdie006 Thank you, good catch!

